PR #33, opened 2023-05-09
feat: enable bootstrap to handle only one class

MartinBernstorff:
2 edits:

* `n_bootstraps`: Required for plots to take less than 10 minutes per plot. Also, acceptable according to:
> "An Introduction to the Bootstrap" (1994) by Efron and Tibshirani, they report that you can get a decent estimate with B=25, and B=200 you approach a similar coefficient of variation as infinity. They provide a table of the coefficients of variation for various B (p. 52-53, both pages are available on google books).

* Added a try/except to `metric_wrapper`
For e.g. `roc_auc_by_hour_of_day`, some of the resampled datasets included only one class for true values (i.e. only negatives), and as such, `roc_auc` was undefined. In those cases, return `np.nan`. Makes for some somewhat ugly figures:
auc_by_hour_of_day

But I don't see how else to fix it. Any suggestions, @KennethEnevoldsen? :-)

### DIFF
--- a/src/psycop/model_evaluation/binary/utils.py
+++ b/src/psycop/model_evaluation/binary/utils.py
@@ -11,7 +11,7 @@ def calc_performance(
     df: pd.DataFrame,
     metric: Callable,
     confidence_interval: Optional[float] = None,
-    n_bootstraps: int = 1000,
+    n_bootstraps: int = 100,
     **kwargs: Any,
 ) -> pd.Series:
     """Calculates performance metrics of a df with 'y' and 'input_to_fn' columns.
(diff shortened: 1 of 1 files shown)


MartinBernstorff:
Alternatively, we might drop all the `np.nan`s from the array, and then compute the CI?

KennethEnevoldsen:
Hmmm come to think of it - we are really computing each of them separately, but I suppose they are not really 100% independent (1 follows 2 etc.) there might be a better way. I.e. the real AUC at 6 is probably lower as 5 and 7 are lower. So some kind of dependence could be modelled in.

See e.g. https://james-brennan.github.io/posts/lowess_conf/

PR #51, opened 2023-05-15
refactor: delete shared plots and split up performance by input

MartinBernstorff:
This is huge Lasse, so don't review it. Let's talk about it later today.


src/psycop/common/model_evaluation/binary/time/timedelta_data.py
```diff
@@ -1,58 +1,65 @@
...
-def create_performance_by_timedelta(
+
+def get_timedelta_series(
+    direction: Literal["t1-t2", "t2-t1"],
+    bin_unit: TIMEDELTA_STRINGS,
+    df: pd.DataFrame,
+    t2_col_name: str,
+    t1_col_name: str,
+) -> pd.Series:
+    """Caclulate the time difference between two timestamps."""
```
HLasse, line 27:
```suggestion
    """Calculate the time difference between two timestamps."""
```

HLasse (approved):

PR #61, opened 2023-05-16
fix: align y_hat and y_hat_probs naming

sarakolding:
# Notes for reviewers
The code actually broke for me, because the names weren't aligned. Changed that and added as_index=False in groupby to avoid the bin_unit column becoming an index


MartinBernstorff:
Tests are still failing; let me know what you need from me :-)

src/psycop/common/model_evaluation/binary/subgroup_data.py
```diff
@@ -39,11 +39,20 @@ def get_auroc_by_input_df(
...
+        )
+
+    else:
+        df["n_in_bin"] = df.groupby(input_name)[input_name].transform("count")
+
+        output_df = auroc_by_group(
+            df=df,
+            groupby_col_name=input_name,
+            confidence_interval=confidence_interval,
+        )
```
MartinBernstorff, line 55: I don't understand this; if we're not binning, why does "n_in_bin" make sense? :-)

src/psycop/common/model_evaluation/binary/time/test_timedelta_data.py
```diff
@@ -37,7 +37,7 @@ def test_get_timedelta_series():
 
 def test_sensitivity_by_timedelta():
     df = str_to_df(
-        """t1_timestamp,t2_timestamp,y,pred,
+        """t1_timestamp,t2_timestamp,y,y_hat_probs,
```
MartinBernstorff, line 40:
```suggestion
        """t1_timestamp,t2_timestamp,y,y_hat,
```

For sensitivity, we need the prediction, not the score :-)

src/psycop/common/model_evaluation/binary/time/timedelta_data.py
```diff
@@ -149,7 +149,7 @@ def get_auroc_by_timedelta_df(
 
 def get_sensitivity_by_timedelta_df(
     y: Iterable[int],
-    y_pred: Iterable[float],
+    y_hat_probs: Iterable[float],
```
MartinBernstorff, line 152:
```suggestion
    y_hat: Iterable[float],
```

src/psycop/common/model_evaluation/binary/time/timedelta_data.py
```diff
@@ -165,7 +165,7 @@ def get_sensitivity_by_timedelta_df(
 
     Args:
         y: True labels
-        y_pred: The predicted class.
+        y_hat_probs: The predicted class.
```
MartinBernstorff, line 168:
```suggestion
        y_hat: The predicted class.
```

src/psycop/common/model_evaluation/binary/time/timedelta_data.py
```diff
@@ -182,7 +182,7 @@ def get_sensitivity_by_timedelta_df(
     """
     df = get_timedelta_df(
         y=y,
-        y_hat=y_pred,
+        y_hat_probs=y_hat_probs,
```
MartinBernstorff, line 185:
```suggestion
        y_hat_probs=y_hat_probs,
```
Maybe generalise the name? E.g. "prediction" or something, which is then just carried over?

src/psycop/common/model_evaluation/binary/time/timedelta_data.py
```diff
@@ -221,14 +221,17 @@ def create_sensitivity_by_time_to_outcome_df(
         pd.DataFrame
     """
 
-    y_hat_series, actual_positive_rate = eval_dataset.get_predictions_for_positive_rate(
+    (
+        y_hat_probs_series,
```
MartinBernstorff, line 225: Rename as above

MartinBernstorff (approved):
Thanks for this! A few edits, then it looks good 👍

src/psycop/common/model_evaluation/binary/subgroup_data.py
sarakolding: n_in_group instead? because we are grouping

sarakolding: and because i would like the number in groups for plotting (sex)

src/psycop/common/model_evaluation/binary/time/timedelta_data.py
sarakolding: i don't understand 😢

sarakolding:
@MartinBernstorff i think i just realised why i'm having so much trouble getting the naming to fit across all functions. get_timedelta_df gets used for both downstream auroc and sensitivity functions, where the columns is ofc named y_hat_probs and y_hat, respectively. i don't see how i can get it to fit both

sarakolding:
> @MartinBernstorff i think i just realised why i'm having so much trouble getting the naming to fit across all functions. get_timedelta_df gets used for both downstream auroc and sensitivity functions, where the columns is ofc named y_hat_probs and y_hat, respectively. i don't see how i can get it to fit both, but it might be exam brain, so let me know if you can see a way to fix it

ah, this was maybe what you meant above. will try out prediction

sarakolding:
nope, still breaks in auc_roc_by_group, because this function is sometimes fed data from get_timedelta_df (y_pred) and sometimes from other functions, e.g. get_auroc_by_input_df (y_hat_probs)

sarakolding:
> nope, still breaks in auc_roc_by_group, because this function is sometimes fed data from get_timedelta_df (y_pred) and sometimes from other functions, e.g. get_auroc_by_input_df (y_hat_probs)

should i just rename the columns in some functions?

MartinBernstorff:
Ah, I see! When it's used for AUROC calculations, it needs to return `y_hat_probs`. When used for sensitivity, we need `y_hat`.

In `get_timedelta_df`, we probably want to rename `y_hat` to something like `output`, and then document it well in the docstring. The only think `y_hat` is used for there is a join. Then `get_auroc_by_timedelta_df` and `get_sensitivity_by_timedelta_df` can rename the `output` column to `y_hat_probs` and `y_hat`, respectively 👍

PR #75, opened 2023-05-25
mb: t2d full eval of chosen pipeline

MartinBernstorff:
# Notes for reviewers
Exclusively review the code changes in `common`, as well as the two comments I've written below. I'll fix any changes you recommend :-)


src/psycop/projects/t2d/paper_outputs/aggregate_eval/md_objects.py
```diff
@@ -0,0 +1,128 @@
+from collections.abc import Sequence
```
MartinBernstorff, line 1: @HLasse, perhaps we want to add this to common?

src/psycop/projects/t2d/utils/pipeline_objects.py
```diff
@@ -82,62 +114,103 @@ def cfg(self) -> FullConfigSchema:
...
+            create_output_paths_on_init=create_output_paths_on_init,
+        )
+
+
+class PipelineRun:
+    def __init__(
+        self,
+        name: str,
+        group: RunGroup,
+        pos_rate: float,
```
MartinBernstorff, line 191: Perhaps we want to add this abstraction to the library? It's strictly coupled to the folder structure output by `train_model`, so if we want it, we need a contract-test that ensures that the interface between the two is stable.

PR #168, opened 2023-06-28
feat: sequence windower

MartinBernstorff:
Changes to do:
- [ ] Benchmark
- [ ] Validate dataframes on init


psycop/common/feature_generation/sequences/timeseries_windower/test_timeserieswindower.py
```diff
@@ -0,0 +1,211 @@
...
+    return SequenceColumns()
+
+
+class TestTimeserieswindower:
+    @staticmethod
+    def test_windowing_should_be_fast(
+        c: SequenceColumns,
+        n_patients: int = 1000,
+        n_event_dfs: int = 15,
+        events_per_patient: int = 50,
```
HLasse, line 62: benchmark with a lookbehind too

psycop/common/feature_generation/sequences/timeseries_windower/types/prediction_time_dataframe.py
```diff
@@ -0,0 +1,51 @@
...
+
+        if self._cols.pred_time_uuid not in self._df.columns:
+            self._df = self._df.with_columns(
+                create_pred_time_uuids(
+                    entity_id_col_name=self._cols.entity_id,
+                    timestamp_col_name=self._cols.timestamp,
+                ).alias(self._cols.pred_time_uuid),
+            )
+        else:
+            df = self._df
```
HLasse, line 43: is there a reason for having these lines?

psycop/common/feature_generation/sequences/timeseries_windower/timeseries_windower.py
```diff
@@ -0,0 +1,42 @@
...
+    SequenceDataframeBundle,
+)
+
+
+def window_timeseries(
+    prediction_times_bundle: PredictiontimeDataframeBundle,
+    event_bundles: Sequence[EventDataframeBundle],
+    lookbehind: datetime.timedelta | None = None,
+) -> SequenceDataframeBundle:
+    pred_time_df, pred_time_cols = prediction_times_bundle.unpack()
```
HLasse, line 23: This fella could use a docstring as it's the main/only user-facing function

HLasse (approved):
Looks fantastic, Martin! Let's use PolarsDataFrameBUndle for __everything__

psycop/common/feature_generation/sequences/timeseries_windower/types/abstract_polars_dataframe.py
```diff
@@ -0,0 +1,47 @@
...
+class ColumnBundle:
+    entity_id: str = "entity_id"
+
+
+class PolarsDataframeBundle:
+    def __init__(
+        self,
+        df: pl.LazyFrame,
+        cols: ColumnBundle,
+        validate_cols_exist_on_init: bool = True,
```
KennethEnevoldsen, line 17: Why would this be optional? (is it due to being lazy?)

psycop/common/feature_generation/sequences/timeseries_windower/types/event_dataframe.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+@dataclass(frozen=True)
+class EventColumns(ColumnBundle):
+    timestamp: str = "event_timestamp"
+    event_source: str = "event_source"  # Values are e.g. "lab"/"diagnosis"/"medication"
+    event_type: str = "event_type"  # Values are e.g. "Hba1c"/"Hypertension"/"Metformin"
+    event_value: str = (
+        "event_value"  # 1/0 for booleans, numeric value for numeric events
+    )
```
KennethEnevoldsen, line 18: Initially misunderstood this object. I would document what a columnbundle is.

KennethEnevoldsen (commented):
Looks goods, but the test are a bit unclear I would like to review it again with a set of updated tests. If we assume that data frames is the right abstraction I think the abstractions looks meaningful.

The following is a meta consideration that I think worth noting, but probably not be worth implementing without considerable considerations as it probably is a complete rethinking of major parts of the codebase.

I think these abstractions are already pointing in a direction where you could define the completely agnostic to a data frame frame  (Event, Outcome, etc.), and then simply built the data frame at the moment before inference.

This would e.g. remove all the handling of Columns names and a lot (all?) of the merge operations (which is what I understand is the primary reason for the DF abstraction) becomes trivial once  at the right layer of abstraction (there is no need to merge on PatientID, is we simply have a patient object). It would also alleviate dataset splitting.

MartinBernstorff:
> Looks goods, but the test are a bit unclear I would like to review it again with a set of updated tests. If we assume that data frames is the right abstraction I think the abstractions looks meaningful.
> 
> The following is a meta consideration that I think worth noting, but probably not be worth implementing without considerable considerations as it probably is a complete rethinking of major parts of the codebase.
> 
> I think these abstractions are already pointing in a direction where you could define the completely agnostic to a data frame frame (Event, Outcome, etc.), and then simply built the data frame at the moment before inference.
> 
> This would e.g. remove all the handling of Columns names and a lot (all?) of the merge operations (which is what I understand is the primary reason for the DF abstraction) becomes trivial once at the right layer of abstraction (there is no need to merge on PatientID, is we simply have a patient object). It would also alleviate dataset splitting.

Re: Level of abstraction, I take it that you'd want to use pure python and objects (or something similar?) to represent the data? I would love to do this, because it unlocks simple static type checking (like you alluded to).

Lasse and I considered it as well, but decided against it, mainly from a performance stand-point. Since we need to do a many-to-many join (each "event", e.g. blood-sample, can be part of the sequence for many prediction times), choosing an implementation that is easily parallelised appeared to be a good move. 

If we chose pure python, this join wouldn't be represented as a join, but in some other way - e.g. a for loop that adds each event to all relevant prediction times. Either way, we'd have to do an operation that is functionally equivalent to the join. My expectation is that this would be much slower if implemented in pure python or numpy arrays.

Would love a chat about this at some point! Looking at the PR now.

psycop/common/feature_generation/sequences/timeseries_windower/types/abstract_polars_dataframe.py
MartinBernstorff: Makes for easier testing, e.g. in `test_should_be_frozen`.

MartinBernstorff:
> > I think these abstractions are already pointing in a direction where you could define the completely agnostic to a data frame frame (Event, Outcome, etc.), and then simply built the data frame at the moment before inference.
> > This would e.g. remove all the handling of Columns names and a lot (all?) of the merge operations (which is what I understand is the primary reason for the DF abstraction) becomes trivial once at the right layer of abstraction (there is no need to merge on PatientID, is we simply have a patient object). It would also alleviate dataset splitting.
> 
> Re: Level of abstraction, I take it that you'd want to use pure python and objects (or something similar?) to represent the data? I would love to do this, because it unlocks simple static type checking (like you alluded to).
> 
> Lasse and I considered it as well, but decided against it, mainly from a performance stand-point. Since we need to do a many-to-many join (each "event", e.g. blood-sample, can be part of the sequence for many prediction times), choosing an implementation that is easily parallelised appeared to be a good move.
> 
> If we chose pure python, this join wouldn't be represented as a join, but in some other way - e.g. a for loop that adds each event to all relevant prediction times. Either way, we'd have to do an operation that is functionally equivalent to the join. My expectation is that this would be much slower if implemented in pure python or numpy arrays.
> 
> Would love a chat about this at some point! Looking at the other comments now.

FEMR have done it in pure python: https://github.com/som-shahlab/femr/blob/main/tutorials/1_


(The authors of the MOTOR paper, doing survival analysis with deep learning): [Steinberg et al. - 2023 - Self-Supervised Time-to-Event Modeling with Struct.pdf](https://github.com/Aarhus-Psychiatry-Research/psycop-common/files/11914732/Steinberg.et.al.-.2023.-.Self-Supervised.Time-to-Event.Modeling.with.Struct.pdf)
Overview.ipynb

psycop/common/feature_generation/sequences/timeseries_windower/types/abstract_polars_dataframe.py
KennethEnevoldsen: Virker som et funky trade-off? Hvorfor ikke bare have rigtige kolonner i testen?

KennethEnevoldsen:
> Lasse and I considered it as well, but decided against it, mainly from a performance stand-point. Since we need to do a many-to-many join (each "event", e.g. blood-sample, can be part of the sequence for many prediction times), choosing an implementation that is easily parallelised appeared to be a good move.

I don't believe it would be slower. I actually think you can completely avoid the join. You can simply examine the events you have in your lookback window and construct the desired dataframe pr. patient and then the merge is a (fast) concat.

> Either way, we'd have to do an operation that is functionally equivalent to the join.

Provide the same result I agree, but I do believe you could obtain it faster or equivalently using the other abstraction (You have a lot of additional assumptions about the data which you can likely utillize)

> choosing an implementation that is easily parallelised appeared to be a good move.
I think what you mean to say is that we have access to optimized rust implementations (the other approach is just as parallizable) and I do get that, but I am not sure it results in a net gain (it might).

--- 
Anyway argument was that it was possible. I am not sure if it is worth it atm. 

I do find the abstraction of the FEMR package a lot more pleasant to read (and likely debug).

KennethEnevoldsen (approved):

MartinBernstorff:
> I don't believe it would be slower. I actually think you can completely avoid the join. You can simply examine the events you have in your lookback window and construct the desired dataframe pr. patient and then the merge is a (fast) concat.

Let's riff on this! So, the way I see it, this means we'd:

1. Create an object for each patient (join + groupby)
2. Bucket-sort events to each patient (join + groupby)
3. Sort the events within patients (filter)
4. Slice the event-series at the correct lookbehind distance (filter)

I'd imagine that's basically what's happening under the hood in polars anywho. 

That said, the implementation is super fast, so I'd love to see a test of it! (Or just have a chat about it). 

I'll merge this for now, with the caveat that we're strongly considering another abstraction.

MartinBernstorff:
@KennethEnevoldsen I've experimented with it a bit here: https://github.com/Aarhus-Psychiatry-Research/timeseriesflattener/blob/fb71e35147e909ab1d1bdbb0146ea4ba2693906c/src/timeseriesflattener/tests/test_timeseriesflattener/test_flattened_dataset/test_add_values.py

Took 6 seconds to unpack 5.000 patients with 10 prediction times and 10 predictor values each. That's after the dataframe is sorted by `entity_id` and in memory as dicts.

For T2D, it's 120.000 patients with ~20 visits each and ~30 predictor values. Assuming best-case (linear time), this makes it 6 seconds · 24x patients · 2x visits = 288 seconds for unpacking. 

That's not too bad! Now, we're lacking a multiplier for the number of events, as well as profiling the time spent sorting the dataframe. I just tested with 120.000 patients and 10 events per patient, that dataframe took ~24 seconds to sort and move out into dicts.

There's also quite possibly some performance benefit to get by refining the loop. What do you think?

PR #192, opened 2023-08-04
added function to create sparse_feature_set

erikperfalk:
# Notes for reviewers
Is there a better name than sparse_feature_set? I avoided baseline/base as this could maybe be confused with "baseline models".


bokajgd (approved):
Ser fint ud! Jeg merger. :)

PR #209, opened 2023-08-22
feat: move feature_set_dir check to earlier 

frillecode:
# Notes for reviewers
As discussed. 
Moved check for whether you want to overwrite files in the `feature_set_dir` path. Now it is checked before the feature generation is executed so that we will not potentially generate features that we would not want to save


bokajgd (approved):
Ser bare fint ud!

PR #217, opened 2023-08-24
Sk/classifierchain

sarakolding:
# Notes for reviewers
Changed multioutput to classifierchain for multilabel classification. However, along the way found a bug in ArtifactsToDiskSaver, which was caused by the changes I made in the EvalDataset, to accomodate multiple labels (cannot highlight here, since the code was changed in the previous pr, not this one, but I changed the y and y_hat_probs attribute in the EvalDataset to "pd.Series | pd.DataFrame" instead of pd.Series). It results in these labels not being saved in the evaluation_dataset.parquet file if they are in dataframe format. Did a quick hack to unpack them so it works again, but not optimal. Maybe I should write a test of whether EvalDataset contains something in all non-custom attributes, to avoid in future? 
Got a bit confused about the utility of the EvalDataset class along the way. Seems we often end up transforming the data back to dataframe format for eval?


bokajgd (commented):
Code looks great! Testing that the eval_datset columns always hold values before exporting to disk might be a good idea so we catch potential breaks in the future. Otherwise, stuff like this might go unnoticed for quite a while and become confusion.

HLasse:
@sarakolding do you need me to review this as well? :)

sarakolding:
> @sarakolding do you need me to review this as well? :)

Think I solved the mess with the BEHRT branch, and added some tests. Would you maybe look over those, @HLasse? 👼

PR #229, opened 2023-09-26
feat: time from first positive prediction to next hba1c

MartinBernstorff:
# Notes for reviewers


psycop/projects/t2d/paper_outputs/intervention_eval/hba1c.py
```diff
@@ -0,0 +1,79 @@
...
+            hba1cs,
+            on="patient_id",
+            how="inner",
+        )
+        .with_columns(
+            (pl.col("timestamp") - pl.col(first_pos_pred_colname)).alias(
+                delta_time_col_name,
+            ),
+        )
+        .filter(pl.col(delta_time_col_name) >= 0)
```
sarakolding, line 31: what about test of this line? (prediction times before first tp should be excluded)

sarakolding (changes requested):

sarakolding (approved):
good job :-)

PR #238, opened 2023-10-02
ci: add dev container

MartinBernstorff:
Adding a development container as an option; should be especially helpful when e.g. @sarakolding gets her new mac.

@HLasse @KennethEnevoldsen Just wanted to show this off - let me know if you want to chat about it 👍

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -18,8 +18,7 @@ RUN curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | d
 
 # Install dependencies
 COPY dev-requirements.txt requirements.txt test-requirements.txt .
-# RUN pip install -r dev-requirements.txt
-RUN pip install pre-commit==3.4.0 
+RUN pip install -r dev-requirements.txt
 
 # Install the app
 COPY . /app
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,20 +1,30 @@
 # Use Python 3.10 slim as the base image
 FROM python:3.10
-RUN apt-get update && apt-get install -y git
+RUN apt-get update && apt-get install -y git-all
 
 # Set the working directory to /app
 WORKDIR /app
 
+# Lazygit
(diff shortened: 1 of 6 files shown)


Dockerfile.dev
```diff
@@ -0,0 +1,36 @@
...
+# Upgrade pip
+RUN pip install --upgrade pip
+
+# Set the working directory to /app
+WORKDIR /app
+VOLUME psycop-common
+
+# Dev experience
+## Lazygit
+RUN LAZYGIT_VERSION=$(curl -s "https://api.github.com/repos/jesseduffield/lazygit/releases/latest" | grep -Po '"tag_name": "v\K[^"]*') && curl -Lo lazygit.tar.gz "https://github.com/jesseduffield/lazygit/releases/latest/download/lazygit_${LAZYGIT_VERSION}_Linux_x86_64.tar.gz" && tar xf lazygit.tar.gz lazygit && install lazygit /usr/local/bin && rm -rf lazygit lazygit.tar.gzRUN curl https://raw.githubusercontent.com/jesseduffield/lazydocker/master/scripts/install_update_linux.sh | bash
```
KennethEnevoldsen, line 14: what does this do?

Dockerfile.dev
```diff
@@ -0,0 +1,36 @@
...
+COPY .pre-commit-config.yaml ./
+RUN git init . && pre-commit run
+
+# App install
+## Each on their own line to have them be their own layers, optimising caching
+COPY requirements.txt .
+RUN pip install -r requirements.txt
+
+COPY test-requirements.txt .
+RUN pip install -r test-requirements.txt
```
KennethEnevoldsen, line 34: having multiple requirements files could potentially lead to problems (e.g. if a test-req. has a requirement which is incompatible with a req. you will just overwrite the req.). I would just convert it all to one requirements file (potentially by just concatenating them in the script or commenting sections)

Dockerfile.dev
MartinBernstorff: Install a command-line utility called lazygit.

https://github.com/jesseduffield/lazygit

PR #255, opened 2023-10-10
frillecode/cast-diagnoses-to-caliber

frillecode:
# Notes for reviewers


psycop/common/feature_generation/sequences/mapping_diagnosis_codes.py
frillecode, line 1: Could just delete the ```extract_icd10_to_caliber_categories_df()``` as we do not need it anymore. It was just used to create the .csv file with the mapping

PR #260, opened 2023-10-10
feat: filter diagnosis subtype in BEHRT

frillecode:
Added filtering of diagnosis subtypes so we only keep A-diagnoses.

### DIFF
--- a/psycop/common/sequence_models/tests/test_embedding.py
+++ b/psycop/common/sequence_models/tests/test_embedding.py
@@ -49,6 +49,7 @@ def test_diagnosis_mapping(
         date_of_birth=dt.datetime(year=1990, month=1, day=1),
     )
 
+    # Add temporal events to check that diagnosis codes are mapped correctly
     temporal_events = [
         TemporalEvent(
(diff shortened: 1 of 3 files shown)


psycop/common/sequence_models/tests/test_embedding.py
```diff
@@ -54,19 +54,25 @@ def test_diagnosis_mapping(
...
             value="A30",
             source_type="diagnosis",
-            source_subtype="test_name",
+            source_subtype="A",
+        ),
+        TemporalEvent(
+            timestamp=dt.datetime(2021, 1, 3),
+            value="A30",
+            source_type="diagnosis",
+            source_subtype="B",
```
MartinBernstorff, line 69: Would propose adding comments to each of these TemporalEvents, documenting why they're included. 

E.g. I assume this example is included to test that subtype = "B" is excluded? :-)

MartinBernstorff (commented):
Looks good! Only one comment, then ready to merge for me 👍

MartinBernstorff:
We could also use an enum for source subtype; this is exactly their use-case 👍

PR #279, opened 2023-10-19
feat: allow levels of granularity in diagnosis mapping

frillecode:
Noticed that we lose a lot of diagnoses when mapping from ICD-10 to CALIBER (~40% of patients have empty sequences after mapping).

CALIBER uses ICD-10 with three or four characters (e.g. "A20" and "A201").

Before, get_mapped_diagnoses_codes() matches the codes directly and if they do not exist in the mapping, it excludes them. Changed this so that it checks different levels of granularity, starting with the highest. (e.g. for "A201" it first checks for "A201" in the mapping and if that does not exist it checks for "A20" and then uses that, otherwise excludes it).

Now, only ~10% of patients have empty sequences after mapping.

### DIFF
--- a/psycop/common/sequence_models/embedders/BEHRT_embedders.py
+++ b/psycop/common/sequence_models/embedders/BEHRT_embedders.py
@@ -199,7 +199,17 @@ class BEHRTEmbedder(nn.Module):
         ) as fp:
             mapping = json.load(fp)
 
-        return [mapping[d] for d in diagnosis_codes if d in mapping]
+        # For each diagnosis code, attempt to map to caliber code
+        # If no mapping exists, remove one character from the end of the code and try again
+        mapped_diagnosis_codes = []
+        for d in diagnosis_codes:
+            while len(d) > 2:  # only attempt codes with at least 3 characters
+                if d in mapping:
+                    mapped_diagnosis_codes.append(mapping[d])
+                    break
+                d = d[:-1]  # noqa: PLW2901
+
+        return mapped_diagnosis_codes
 
     def collate_patient(self, patient: Patient) -> dict[str, torch.Tensor]:
         events = patient.temporal_events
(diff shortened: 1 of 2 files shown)


MartinBernstorff:
A sidenote - perhaps we should raise an error or warning, if we are generating PatientSlices with no contents?

psycop/common/sequence_models/embedders/BEHRT_embedders.py
```diff
@@ -199,7 +199,17 @@ def get_mapped_diagnosis_codes(
         ) as fp:
             mapping = json.load(fp)
 
-        return [mapping[d] for d in diagnosis_codes if d in mapping]
+        # For each diagnosis code, attempt to map to caliber code
```
MartinBernstorff, line 202: A tiny nitpick. The method is not a 'getter' because it transforms the value. I'd rename to something like `map_icd_to_calibre` :-)

psycop/common/sequence_models/tests/test_embedding.py
```diff
@@ -76,10 +76,10 @@ def test_diagnosis_mapping(
             source_type="diagnosis",
             source_subtype="A",
         ),
-        # Check that diagnosis codes that are not in the mapping are excluded
```
MartinBernstorff, line 79: We should still test exclusion :-)

MartinBernstorff (approved):
Tiny change requested, then ready to go! Looks great!

psycop/common/sequence_models/tests/test_embedding.py
frillecode: This is already tested as the```Patient(
        patient_id=11,
        date_of_birth=dt.datetime(year=1990, month=1, day=1),
    ) ``` has temporal events with values "d1" or "d2" (which are not in the mapping). 
  
Adding it here would be redundant but maybe make it more clear that we are testing it?

PR #285, opened 2023-10-23
Speed up tests

HLasse:
Subsample test datasets even further


HLasse:
Tests passed 4 times on my PC (and on CI) so seem to be stable

PR #349, opened 2023-11-06
implemented binary classification pipeline

HLasse:
# Notes for reviewers


psycop/common/model_training_v2/classifier_pipelines/binary_classification_pipeline.py
```diff
@@ -10,10 +11,19 @@
...
-        ...
+        self.pipe = Pipeline(steps=steps)
 
-    def fit(self, X: PolarsFrame, y: PolarsFrame) -> None:
-        ...
+    def fit(self, X: PolarsFrame, y: pl.Series) -> None:
+        if isinstance(X, pl.LazyFrame):
+            # pyright thinks "X" is a constant since it's all caps
+            # hence we type ignore
+            X = X.collect()  # type: ignore
```
MartinBernstorff, line 20: Maybe we should not call it big-X then? Just small-x?

psycop/common/model_training_v2/classifier_pipelines/binary_classification_pipeline.py
```diff
@@ -10,10 +11,19 @@
...
+        self.pipe.fit(X.to_pandas(), y)
 
     def predict_proba(self, X: PolarsFrame) -> pl.Series:
-        ...
+        """Returns the predicted probabilities of the `1`
+        class"""
+        if isinstance(X, pl.LazyFrame):
+            X = X.collect()  # type: ignore
+        pred_probs = self.pipe.predict_proba(X.to_pandas())[:, 1]
+        return pl.Series("y_hat_probs", pred_probs)
```
MartinBernstorff, line 29: Would be neat if `y_hat_probs` this could be type-hinted in the return type. E.g. the return type is a "PredProbaSeries" or similar.

MartinBernstorff (approved):
Looks good! Small suggestions.

psycop/common/model_training_v2/classifier_pipelines/binary_classification_pipeline.py
MartinBernstorff: Then we can remove the type ignore.

HLasse:
We have a type issue here. If we type hint the type of a pd.Series with e.g. `pd.Series[int]` it won't work during runtime! However, if we don't type hint the generic type of the series, we get a type-error with pyright in strict mode. Seems to be a known issue https://github.com/microsoft/pylance-release/issues/3445 - not entirely sure what the best way around this is..

PR #356, opened 2023-11-06
Eval dataset & binary classification

HLasse:
# Notes for reviewers


psycop/common/model_training_v2/metrics/binary_metrics/base_binary_metric.py
```diff
@@ -4,9 +4,7 @@
 
 from psycop.common.model_training_v2.metrics.base_metric import CalculatedMetric
 
-from ...presplit_preprocessing.polars_frame import PolarsFrame
-
 
 class BinaryMetric(Protocol):
-    def calculate(self, y_true: PolarsFrame, y_pred: pl.Series) -> CalculatedMetric:
+    def calculate(self, y_true: pl.Series, y_pred: pl.Series) -> CalculatedMetric:
```
MartinBernstorff, line 9: I dunno if this is possible, but can we do pl.Series[int]?

Some binary metrics will need a threshold as well (e.g. accuracy); not sure how to handle that. Perhaps not a concern if we only want to support AUROC.

psycop/common/model_training_v2/metrics/binary_metrics/binary_auroc.py
```diff
@@ -5,12 +5,9 @@
...
 )
-from psycop.common.model_training_v2.presplit_preprocessing.polars_frame import (
-    PolarsFrame,
-)
 
 
 class BinaryAUROC(BinaryMetric):
-    def calculate(self, y_true: PolarsFrame, y_pred: pl.Series) -> CalculatedMetric:
+    def calculate(self, y_true: pl.Series, y_pred: pl.Series) -> CalculatedMetric:
         # sklearn expects MatrixLike | ArrayLike which PolarsFrame and pl.Series fulfill but are not typed as
```
MartinBernstorff, line 12: We talked about this a bit before. The type:ignore makes me worried, perhaps metrics should just support pd.Series instead?

psycop/common/model_training_v2/metrics/binary_metrics/base_binary_metric.py
MartinBernstorff: This will be a problem in BaseEvalDataset.calculate_metrics as well.

psycop/common/model_training_v2/pipeline.py
```diff
@@ -23,4 +23,4 @@ def train_baseline_model(cfg: BaselineSchema) -> float:
     result = cfg.training_method.train()
     result.eval_dataset.to_disk(path=cfg.experiment_path)
 
-    return result.metric
+    return result.metrics
```
MartinBernstorff, line 26: TrainingResult doesn't have a .metrics attribute? How come this does not result in a warning?

psycop/common/model_training_v2/problem_type/eval_dataset_base.py
```diff
@@ -4,27 +4,41 @@
...
 @dataclass
 class BaseEvalDataset:
-    pred_time_uuids: pl.Expr
-    y_hat_probs: pl.Expr
-    y: pl.Expr
-    df: PolarsFrame
+    pred_time_uuids: str
+    y_hat_probs: str
+    y: str
+    df: pl.DataFrame
```
MartinBernstorff, line 19: If they are pl.Expr, they can be pl.Col("pred_time_uuids") for example. Might want a comment for that instead?

psycop/common/model_training_v2/problem_type/eval_dataset_base.py
```diff
@@ -4,27 +4,41 @@
...
+                y_true=self.df.get_column(self.y),
+                y_pred=self.df.get_column(self.y_hat_probs),
+            )
+            for metric in metrics
+        ]
 
-    def calculate_metrics(self, metrics: Iterable[BinaryMetric]) -> dict[str, float]:
-        ...
+    def to_disk(self, path: Path) -> None:
+        write_to_pickle(self, filepath=path)
```
MartinBernstorff, line 44: We definitely do not want to pickle this. Will break all the time. Rather, the disk-interface should just be parquet or whatever. Probably want users to decide here.

psycop/common/model_training_v2/problem_type/problem_type_base.py
```diff
@@ -41,9 +47,37 @@ def __init__(
...
         X: PolarsFrame,
-        y: PolarsFrame,
-    ) -> float:
-        ...
+        y: pl.Series,
+    ) -> None:
+        self.pipe.fit(X, y)
+
+    def predict_proba(self, x: PolarsFrame) -> pl.Series:
+        return self.pipe.predict_proba(x)
```
MartinBernstorff, line 55: Won't this return a pd.Series?

psycop/common/model_training_v2/problem_type/problem_type_base.py
```diff
@@ -41,9 +47,37 @@ def __init__(
...
+
+    def evaluate(self, x: PolarsFrame, y: pl.Series) -> TrainingResult:
+        if isinstance(x, pl.LazyFrame):
+            x = x.collect()
+        y_hat_probs = self.pipe.predict_proba(x)
+
+        df = x.with_columns(y_hat_probs, y)
+        eval_dataset = BinaryEvalDataset(
+            pred_time_uuids="pred_time_uuids",  # need to get this column from somewhere!
```
MartinBernstorff, line 64: Agreed, this is still unsolved. Maybe you can look at designing a solution? :-)

psycop/common/model_training_v2/problem_type/problem_type_base.py
```diff
@@ -41,9 +47,37 @@ def __init__(
...
+            y_hat_probs=y_hat_probs.name,
```
MartinBernstorff, line 65: If they are pd.Series, do they have a name attribute?

psycop/common/model_training_v2/pipeline.py
HLasse: it did but was somehow missed

psycop/common/model_training_v2/problem_type/eval_dataset_base.py
HLasse: There doesn't seem to be a nice way of selecting columns using `pl.col` though -- if we use `.select` we get a type error in `metric.calculate` because it expects a series. Using `str` seems to me the most flexible solution at the moment, but open to suggestions

psycop/common/model_training_v2/problem_type/problem_type_base.py
HLasse: yep! the column name is set as .name

PR #357, opened 2023-11-06
Adding filtering to train

KennethEnevoldsen:
Added filtering to train loop


psycop/common/sequence_models/train.py
```diff
@@ -54,10 +54,17 @@ def train(config_path: Path | None = None) -> None:
...
+    std_logger.info("Updating Config")
     flat_config = flatten_nested_dict(config_dict)
     logger.experiment.config.update(flat_config)
 
-    # create dataloader:
+    # filter dataset
+    std_logger.info("Filtering Patients")
+    filter_fn = model.embedding_module.filter_and_reformat_events
+    training_dataset.filter_patients(filter_fn)
+    validation_dataset.filter_patients(filter_fn)
```
KennethEnevoldsen, line 66: We filter patient multiple times (both here and for the embedding module), but it is quite fast so I don't think it is a problem.

psycop/projects/sequence_models/pretrain_mini_behrt.cfg
KennethEnevoldsen, line 1: This is a small script for testing

psycop/projects/sequence_models/pretrain_behrt.cfg
KennethEnevoldsen, line 1: This is the config for BEHRT (it should match, though there is naturally a few things which we do not know.

MartinBernstorff:
@KennethEnevoldsen This is stale, but tests pass. Should we merege? :-)

PR #364, opened 2023-11-06
feat: log calculated metric with SplitTrainer

MartinBernstorff:
- misc.
- feat: log calculated metric


HLasse (approved):
Looks good! Can we add a logger.info method though?

PR #379, opened 2023-11-07
tests: add test for v2 pipeline

MartinBernstorff:
Discovered lacking inheritance from the protocols, which meant I had to fix a bunch of minor types. Other than that, should be good.


psycop/common/model_training_v2/loggers/base_logger.py
```diff
@@ -1,11 +1,12 @@
-from typing import Any, Protocol
+from typing import Any, Protocol, runtime_checkable
 
 import wasabi
 
 from psycop.common.global_utils.config_utils import flatten_nested_dict
 from psycop.common.model_training_v2.metrics.base_metric import CalculatedMetric
 
 
+@runtime_checkable
```
HLasse, line 9: What does this do for us?

psycop/common/model_training_v2/problem_type/binary_classification.py
```diff
@@ -13,14 +13,22 @@
...
     BinaryEvalDataset,
 )
+from psycop.common.model_training_v2.problem_type.problem_type_base import ProblemType
 from psycop.common.model_training_v2.training_method.base_training_method import (
     TrainingResult,
 )
 
-## Have .evaluate only return the main metric and df with pred_time_uuid,y_hat_probs
 
+def polarsframe_to_series(polarsframe: PolarsFrame) -> pl.Series:
```
HLasse, line 22: Neat solution to the issue of requiring different types of y for e.g. multilabel and binary classification. We should remember to write docstrings in e.g. BinaryClassification to specify that `y`should be a dataframe with a single column though

psycop/common/model_training_v2/problem_type/problem_type_base.py
```diff
@@ -17,5 +20,5 @@ def train(self, x: PolarsFrame, y: PolarsFrame) -> None:
     def evaluate(self, x: PolarsFrame, y: PolarsFrame) -> TrainingResult:
         ...
 
-    def predict_proba(self, x: PolarsFrame) -> pl.Series:
+    def predict_proba(self, x: PolarsFrame) -> pd.Series[float]:
```
HLasse, line 23: Unsure we can make this assumption with the return type. Multi-label classification will return an array/dataframe with the same number of columns as the number of output labels. We've already considered the multilabel case in the .train and .evaluate methods of ProblemType (by using PolarsFrame instead of Series for y), so I think we should try to come up with a smart solution here as well

psycop/common/model_training_v2/test_pipeline.py
```diff
@@ -0,0 +1,65 @@
...
+                                     1,1,1,1
+                                     2,1,1,99
+                                     3,1,1,99
+                                     4,0,0,99
+                                     5,0,0,99
+                                     6,0,0,99
+                                     """,
+    )
+    logger = TerminalLogger()
+    schema = BaselineSchema(
```
HLasse, line 46: This seems pretty neat! We are great!

psycop/common/model_training_v2/problem_type/problem_type_base.py
HLasse: We actually already made a `PredProbaSeries` type which we use in BinaryClassification. We are never directly calling .predict_proba -- it's called via `.train` -- so one solution could be to make it private, have it return a PredProbaSeries (where each row could be a list of floats for the multilabel case) and handle the rest in the `.train` method of the ProblemType. Let's talk about it

HLasse (approved):

psycop/common/model_training_v2/loggers/base_logger.py
MartinBernstorff: Required for protocols to be used in pydantic objects; pydantic does runtime type checking, but it is limited with protocols (e.g. does not check that the type of arguments is correct).

psycop/common/model_training_v2/problem_type/binary_classification.py
MartinBernstorff: Completely agree. Alternatively, they should not subclass from the same protocol. I've added an issue.

psycop/common/model_training_v2/problem_type/problem_type_base.py
MartinBernstorff: That sounds like a very good solution. Referencing this in #392.

PR #389, opened 2023-11-08
dev: add quick PR using testmon

MartinBernstorff:
This adds an entrypoint for running tests using https://testmon.org/

> pytest-testmon is a pytest plugin which selects and executes only tests you need to run. It does this by collecting dependencies between tests and all executed code (internally using Coverage.py) and comparing the dependencies against changes. testmon updates its database on each test execution, so it works independently of version control.

This dramatically decreases our test run times, especially for a monorepo.

Potential downsides are false negatives. To avoid this, we still run the full test suite in CI.

I've added it as a separate entrypoint so we can test it, before suggesting it as an integration in everyone's workflow.  This is tracked in #390, so we don't forget.

Since it uses code coverage, it won't catch if you have changed an external file. E.g. if a .cfg file changes, any tests using the .cfg file won't run. I think that is fine as long as we run the full test suite remotely. Alternatively, we could run both locally. First testmon-tests (to get fast feedback), then all tests (to get comprehensive feedback).

@KennethEnevoldsen, @HLasse, what do you guys think? :-)

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -475,9 +475,23 @@ def pr(c: Context, auto_fix: bool = True, create_pr: bool = True):
     static_type_checks(c)
     test(c)
 
+
 @task
-def qtest(c: Context):  # noqa: B006
-    test(c, pytest_args=["psycop", "-rfE", "--failed-first", "-p no:cov", "--disable-warnings", "-q", "--durations=5", "--testmon"])
+def qtest(c: Context):
+    test(
+        c,
+        pytest_args=[
+            "psycop",
+            "-rfE",
+            "--failed-first",
+            "-p no:cov",
+            "--disable-warnings",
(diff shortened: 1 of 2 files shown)


tasks.py
```diff
@@ -476,6 +476,42 @@ def pr(c: Context, auto_fix: bool = True, create_pr: bool = True):
...
+            "-p no:cov",
+            "--disable-warnings",
+            "-q",
+            "--durations=5",
+            "--testmon",
+        ],
+    )
+
+
+# TODO: #390 Make more durable testmon implementation
```
KennethEnevoldsen, line 496: ?

KennethEnevoldsen (approved):

HLasse (approved):

PR #404, opened 2023-11-09
feat: use confection for v2 pipeline

MartinBernstorff:
The meat is in:
`psycop/common/model_training_v2/config/baseline_test_config.cfg`

Then a few refactors in:
`psycop/common/model_training_v2/trainer/split_trainer.py`

And the rest should be minor :-)

@KennethEnevoldsen We're nearing the finish line here! Would love to hear what you think, in particular about `psycop/common/model_training_v2/config/baseline_test_config.cfg`.


psycop/common/model_training_v2/config/baseline_test_config.cfg
```diff
@@ -0,0 +1,44 @@
...
+validation_outcome_col_name = "outcome"
+
+[trainer.training_data]
+@baseline_data = "minimal_test_data"
+
+[trainer.validation_data]
+@baseline_data = "minimal_test_data"
+
+[trainer.logger]
+@baseline_loggers = "terminal_logger"
```
MartinBernstorff, line 19: Would love to interpolate this, but got a recursion error if I did `${logger}`. Any suggestions @KennethEnevoldsen?

psycop/common/model_training_v2/config/baseline_registry.py
```diff
@@ -0,0 +1,25 @@
+import catalogue
+from confection import registry
+
+
+class BaselineRegistry(registry):
+    baseline_loggers = catalogue.create("psycop", "baseline_loggers")
```
MartinBernstorff, line 6: The name-spacing is a bit funky (lots of `baseline_`) to avoid crashing with sequence models. Would love suggestions on how to make it simpler, perhaps another layer of nesting?

psycop/common/model_training_v2/config/config_utils.py
```diff
@@ -0,0 +1,20 @@
...
+
+from psycop.common.model_training_v2.config.baseline_registry import BaselineRegistry
+from psycop.common.model_training_v2.config.baseline_schema import (
+    BaselineSchema,
+)
+from psycop.common.model_training_v2.config.populate_registry import (
+    populate_baseline_registry,
+)
+
+populate_baseline_registry()
```
MartinBernstorff, line 13: Populating the registry right before load seems to be an OK way of doing it, @KennethEnevoldsen?

psycop/common/model_training_v2/test_pipeline.py
```diff
@@ -33,38 +42,60 @@
...
-                                     """,
-    )
     logger = TerminalLogger()
     schema = BaselineSchema(
-        experiment_path=tmpdir,
+        project_info=ProjectInfo(experiment_path=tmpdir),
         logger=logger,
         trainer=SplitTrainer(
-            training_data=training_data,
+            training_data=MinimalTestData(),
```
MartinBernstorff, line 71: @HLasse Confection instantiates all the elements at resolve-time, so having a class made more sense. Perhaps we could make a generic loader which just takes a .parquet path as an argument?

psycop/common/model_training_v2/config/baseline_test_config.cfg
KennethEnevoldsen: Hmm odd. Is it due to imports on parsing the config? (I don't see why parsing the config should lead to a recursion error).

I assume you did it like this:
```
[trainer]
 @baseline_trainers = "split_trainer"
 training_outcome_col_name = "outcome"
 validation_outcome_col_name = "outcome"
 logger = ${logger}
 ```

File: psycop/common/model_training_v2/config/baseline_registry.py (hunk 2 above)
Comment (KennethEnevoldsen) [reply psycop/common/model_training_v2/config/baseline_registry.py:6]: just remove `baseline_`? Or does the catalogue.create("psycop", "loggers") lead to a error?
If that is the case it might be worth doing something like:
catalogue.create("psycop/baseline", "loggers")?

File: psycop/common/model_training_v2/config/baseline_schema.py [hunk 5]
```diff
@@ -0,0 +1,18 @@
...
+
+class ProjectInfo(PSYCOPBaseModel):
+    experiment_path: DirectoryPath  # Experiment_path must be a directory which exists
+
+
+class BaselineSchema(BaseModel):
+    class Config:
+        arbitrary_types_allowed = True
+
+    project_info: ProjectInfo
```
KennethEnevoldsen, line 16:
```suggestion
    project_info: DirectoryPath
```
Simpler

KennethEnevoldsen (approved):
Generally I think it looks pretty good. The `baseline_` in the config is a bit annoying, but otherwise it looks pretty good!

psycop/common/model_training_v2/config/baseline_test_config.cfg
MartinBernstorff: Yep, exactly. Huh, we'll have to debug that, but I don't think it should block this merge :tu

PR #423, opened 2023-11-13
refactor: migrate metric calculation from task to trainer

sarakolding:
# Notes for reviewers


psycop/common/model_training_v2/trainer/task/base_task.py
```diff
@@ -16,9 +16,9 @@ def train(self, x: pd.DataFrame, y: pd.DataFrame, y_col_name: str) -> None:
 
     def evaluate(
```
HLasse, line 17: Perhaps rename to `make_eval_dataset` or something similar

HLasse (commented):
Looks good! I think we should split the evaluate function even further, though. So, rename it to e.g.  `construct_eval_dataset`, and move the functionality for calculating the metric outside of the `EvalDataset`. We would then call the metric function in the outer layer (e.g. CV)

psycop/common/model_training_v2/trainer/task/multilabel_classification/multilabel_metrics/base_multilabel_metric.py
```diff
@@ -1,13 +1,20 @@
...
 from psycop.common.model_training_v2.trainer.task.base_metric import (
     CalculatedMetric,
 )
+from psycop.common.model_training_v2.trainer.task.binary_classification.binary_classification_pipeline import (
+    PredProbaSeries,
+)
 
 
+@runtime_checkable
 class MultilabelMetric(Protocol):
```
MartinBernstorff, line 14: Make subclass of BaseMetric, so does not need to be runtime_checkable

psycop/common/model_training_v2/trainer/task/base_metric.py
```diff
@@ -1,7 +1,24 @@
 from dataclasses import dataclass
+from typing import Protocol, runtime_checkable
+
+import pandas as pd
+
+from psycop.common.model_training_v2.trainer.task.binary_classification.binary_classification_pipeline import (
```
MartinBernstorff, line 6: Move PredProbaSeries to base_metric.

Issue #426, opened 2023-11-14
feat: fill defaults from function signatures into .cfg


MartinBernstorff:
https://thinc.ai/docs/api-config#fill

Allows us to specify defaults in the function signatures, and not relying on them in each .cfg. Makes the code nicer to read IMO.

But can also seem like magic if people just call a 'resolve' function, and suddenly the config is changed. Unsure how to make this best.

This is waiting on confection: https://github.com/explosion/confection/issues/47

When that is solved, we can reopen and merge #48.

MartinBernstorff:
OTOH, if defaults exist, we should use them! I'll add this to the loader of baseschema.

svlandeg:
Let us know whether [v0.1.5a0](https://pypi.org/project/confection/0.1.5a0/) fixes this? 🙏

MartinBernstorff:
Absolutely! Will get on this next week 👍

MartinBernstorff:
@svlandeg Seems fixed in #638! Waiting for review to merge.

PR #438, opened 2023-11-15
feat: fill configs and write to disk on load

MartinBernstorff:


MartinBernstorff:
Breaking because of unknown issue with `*args`, issue here: https://github.com/explosion/confection/issues/47

@KennethEnevoldsen, do you have experience here? ❤️

KennethEnevoldsen:
I would run the debugger and check what the type actually is. 

Btw. the issue on confection is quite hard to follow. I would create a minimal example.

MartinBernstorff:
The type is a dict shaped like:
```
{'age_filter': {'@preprocessing': 'age_filter', 'min_age': 4, 'age_col_name': 'pred_age'}}}
pydantic.error_wrappers.ValidationError: 1 validation error for ArgModel
```

MartinBernstorff:
Updated the Confection issue with an example :-)

KennethEnevoldsen:
Hmm odd. You can naturally turn off the type-checking (validate=False). Seem like confection resolved in during [filling](https://github.com/explosion/confection/blob/b0e19dade467a57e19efa616ce96d8a405604f05/confection/__init__.py#L762). However, it seems like it shouldn't be an error as it simply resolve it correctly. Might be that there is a bug in either the config or the resolving.

MartinBernstorff:
Yeah, that succeeds on filling, but removes the `*` key as well! Have added more context to the issue.

KennethEnevoldsen:
Added a few additional stuff to the other issues. It seems like it might be a bug.

MartinBernstorff:
I'll close this for now. We can track it in #426.

PR #441, opened 2023-11-15
feat: add column validator

MartinBernstorff:
@bokajgd @sarakolding I've added both of you as reviewers, as I think adding preprocessing steps is something you will both end up doing. Would love to hear your thoughts!


psycop/common/model_training_v2/trainer/preprocessing/steps/column_validator.py
```diff
@@ -0,0 +1,51 @@
...
+
+        return input_df
+
+    def _column_name_exists(
+        self,
+        column_name: str,
+        df: pl.DataFrame,
+    ) -> MissingColumnError | None:
+        if column_name not in df.columns:
+            return MissingColumnError(column_name=column_name)
```
MartinBernstorff, line 49: @HLasse Tried some more complicated error handling, you might enjoy it!

sarakolding (approved):
i like it picasso

PR #469, opened 2023-11-21
 feat: add vertical concatinator to common

MartinBernstorff:


MartinBernstorff:
It seems I don't get an error here for missing config as well when running `test_registered_functions`. This is kind of weird.

PR #472, opened 2023-11-21
feat: add RegexColumnBlacklist to common

MartinBernstorff:
Also refactors filters into column and row filters.


HLasse (approved):
Lovely!

PR #478, opened 2023-11-23
feat: add FilterWithinSubset

MartinBernstorff:
Auto-created


HLasse (approved):
Looks good. Some changes need to be made in the cross-validation trainer for this to work properly, as the group_id will be dropped if one filters by the `pred_` prefix. Did you and Sara already talk about this or should we create an issue?

PR #487, opened 2023-11-24
fix: drop entity id column during stratified crossval

sarakolding:
# Notes for reviewers


MartinBernstorff (approved):
Looks good 👍

PR #490, opened 2023-11-27
fix: Updated pre-commit hooks

### LINKED ISSUE #489: Update ruff to prevent autofix leading to potential errors
Older versions of ruff auto fix can lead to issues such as:

https://github.com/astral-sh/ruff/issues/1852

which can lead to unintended errors which are hard to debug (had an issue on one of my own projects).

KennethEnevoldsen:
# Notes for reviewers
Reviewers should pay attention to the changed pre-commit file (the rest is simply fixing issues).

pre-commit updated using
```
pre-commit autoupdate
```

and fixed files using:
```
pre-commit run --all-files
```

fixes #489

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@ default_stages: [commit]
 
 repos:
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.1.6
+    rev: v0.1.5
     hooks:
       - id: ruff
         args:
@@ -17,7 +17,7 @@ repos:
           ]
 
   - repo: https://github.com/psf/black
-    rev: 23.11.0
+    rev: 22.6.0
     hooks:
       - id: black
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@ default_stages: [commit]
 
 repos:
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.0.272
+    rev: v0.1.6
     hooks:
       - id: ruff
         args:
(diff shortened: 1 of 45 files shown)


sarakolding (approved):
Looks good to me. I'm a little confused about how the invoke commands relate to the makefile ones, and why we need similar commands in both?

.pre-commit-config.yaml
```diff
@@ -2,7 +2,7 @@ default_stages: [commit]
 
 repos:
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.0.272
+    rev: v0.1.6
```
MartinBernstorff, line 5:
```suggestion
    rev: v0.1.5
```

.pre-commit-config.yaml
```diff
@@ -17,11 +17,11 @@ repos:
           ]
 
   - repo: https://github.com/psf/black
-    rev: 23.3.0
+    rev: 23.11.0
```
MartinBernstorff, line 20:
```suggestion
    rev: 22.6.0
```
Aligning with dev-requirements.txt. I'm fine with alignment the other way around as well 👍

Makefile
```diff
@@ -25,6 +25,10 @@ grow:
 	make create-random-branch
 	make merge-main
 
+lint:
+	@echo "\n––– Linting –––"
+	pre-commit run --all-files
```
MartinBernstorff, line 30: I've been using the makefile for experiments before rolling them out to our shared ops.

`tasks.py` is where our shared commands have been running until now, so I think this should be in `tasks.py` :-)

.pre-commit-config.yaml
MartinBernstorff: I couldn't see from the linked thread which version ruff fixes this in. Feel free to go for 0.1.6 if aligned with dev-requirements.txt

Makefile
MartinBernstorff: This delineation is insufficiently documented, likely because it hasn't been relevant until now. We definitely want to remidy that.

Why keep invoke at all? Ovartaci.

MartinBernstorff: Tracked in #491.

KennethEnevoldsen:
Yea @sarakolding we should make it more clear. Martin made an issue #491 .

PR #493, opened 2023-11-27
feat: add snyk workflow

sarakolding:
snyk not part of dependencies?

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -57,13 +57,13 @@ jobs:
           message: |
             Snyk successfully passed 🐕
             
-        - uses: mshick/add-pr-comment@v2
-        if: ${{ steps.snyk.outputs.snyk_failed == 1 && github.event_name == 'pull_request' }}
-        id: fail_comment
-        with:
-          message-id: ${{ steps.find_comment.outputs.comment-id }}
-          message: |
-            Looks like Snyk failed. Please check the logs for more details.
+        # - uses: mshick/add-pr-comment@v2
+        # if: ${{ steps.snyk.outputs.snyk_failed == 1 && github.event_name == 'pull_request' }}
+        # id: fail_comment
+        # with:
+        #   message-id: ${{ steps.find_comment.outputs.comment-id }}
+        #   message: |
+        #     Looks like Snyk failed. Please check the logs for more details.
 
       - name: Fail workflow
         if: ${{ steps.snyk.outputs.snyk_failed == 1 && github.event_name == 'pull_request' }}
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -31,8 +31,8 @@ jobs:
           pip install invoke
           pip install -r requirements.txt -r test-requirements.txt -r dev-requirements.txt -r gpu-requirements.txt
 
-      # - name: Install snyk
-      #   run: npm install -g snyk
+      - name: Install snyk
+        run: npm install -g snyk
 
       - name: Run Snyk to check for vulnerabilities
         id: snyk
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -25,8 +25,8 @@ jobs:
         with:
           python-version: ${{ matrix.python-version}}
 
-      # - name: Install snyk
-      #   run: pip install snyk
+      - name: Install snyk
+        run: npm install -g snyk
 
       - name: Run Snyk to check for vulnerabilities
         id: snyk
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -25,11 +25,14 @@ jobs:
         with:
           python-version: ${{ matrix.python-version}}
 
-      - name: Install invoke
-        run: pip install invoke==2.1.0
+      - name: Install dependencies
+        shell: bash
+        run: |
+          pip install invoke
+          pip install -r requirements.txt -r test-requirements.txt -r dev-requirements.txt -r gpu-requirements.txt
 
-      - name: Install snyk
-        run: npm install -g snyk
+      # - name: Install snyk
+      #   run: npm install -g snyk
 
       - name: Run Snyk to check for vulnerabilities
         id: snyk
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -25,6 +25,9 @@ jobs:
         with:
           python-version: ${{ matrix.python-version}}
 
+      - name: Install invoke
+        run: pip install invoke==2.1.0
+
       - name: Install snyk
         run: npm install -g snyk
 
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -48,23 +48,3 @@ jobs:
             echo "snyk_failed=1" >> $GITHUB_OUTPUT
             exit 1
           fi
-
-      - uses: mshick/add-pr-comment@v2
-        if: ${{ steps.snyk.outputs.snyk_failed == 1 && github.event_name == 'pull_request' }}
-        id: fail_comment
-        with:
-          message-id: ${{ steps.find_comment.outputs.comment-id }}
-          message: |
-            Looks like Snyk detected vulnerabilities 🐕 Please check the logs for more details.
-            
-        # - uses: mshick/add-pr-comment@v2
-        # if: ${{ steps.snyk.outputs.snyk_failed == 1 && github.event_name == 'pull_request' }}
-        # id: fail_comment
-        # with:
-        #   message-id: ${{ steps.find_comment.outputs.comment-id }}
-        #   message: |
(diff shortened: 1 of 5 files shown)


MartinBernstorff:
Looks great Sara!

We added the comments to provide context when CI failed. Since we are a small group with relatively low turnover, I don't think they are worth the complexity (as @KennethEnevoldsen suggested), so feel free to remove them 👍 

(I know that's super annoying after you've worked hard on adding them!)

`inv snyk` was added quite rapidly as well; I'd suggest adding something like `for requirements_file in Path(".").rglob(".+req.+\.txt")` to make it robust to further additions of requirements as well 👍

sarakolding:
> Looks great Sara!
> 
> We added the comments to provide context when CI failed. Since we are a small group with relatively low turnover, I don't think they are worth the complexity (as @KennethEnevoldsen suggested), so feel free to remove them.
> 
> (I know that's super annoying after you've worked hard on adding them!)

That's fine, was just playing around with it a bit, and I agree --> let's delete 🐒

> `inv snyk` was added quite rapidly as well; currently, it requires hard-coding each requirements.txt path. If we add another one in the future, it won't be picked up. You can improve upon it with some regex and `rglob` if you'd like! 👍

I was going to do this but saw that we have hardcoded the names of the different requirement files in other workflows, so thought it was maybe better to keep it consistent?

sarakolding:
> > Looks great Sara!
> > We added the comments to provide context when CI failed. Since we are a small group with relatively low turnover, I don't think they are worth the complexity (as @KennethEnevoldsen suggested), so feel free to remove them.
> > (I know that's super annoying after you've worked hard on adding them!)
> 
> That's fine, was just playing around with it a bit, and I agree --> let's delete 🐒
> 
> > `inv snyk` was added quite rapidly as well; currently, it requires hard-coding each requirements.txt path. If we add another one in the future, it won't be picked up. You can improve upon it with some regex and `rglob` if you'd like! 👍
> 
> I was going to do this but saw that we have hardcoded the names of the different requirement files in other workflows, so thought it was maybe better to keep it consistent?

So, I deleted the context messages and Lasse and I agreed that we would keep the hardcoded file paths, to keep things consistent. Also added an error message if the existing files do not match what the code specifies. Have tested that the workflow also fails if there is a vulnerability.

.github/workflows/snyk.yml
```diff
@@ -0,0 +1,50 @@
...
+        continue-on-error: false
+        run: |
+          if inv snyk; then
+            echo "Snyk check passed"
+            echo "snyk_failed=0" >> $GITHUB_OUTPUT
+          else
+            echo "Snyk check failed"
+            echo "snyk_failed=1" >> $GITHUB_OUTPUT
+            exit 1
+          fi
```
KennethEnevoldsen, line 50: I assume this should be enough?
```suggestion
        run: |
          inv snyk
```

I would probably convert the command to inv check-packages (in case we switch out snyk).

.github/workflows/snyk.yml
```diff
@@ -0,0 +1,50 @@
...
+          pip install invoke
+          pip install -r requirements.txt -r test-requirements.txt -r dev-requirements.txt -r gpu-requirements.txt
+
+      - name: Install snyk
+        run: npm install -g snyk
+
+      - name: Run Snyk to check for vulnerabilities
+        id: snyk
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
```
KennethEnevoldsen, line 40: Do we need a token for snyk?

.github/workflows/snyk.yml
KennethEnevoldsen: inv scan-for-vulnerabities

.github/workflows/snyk.yml
```diff
@@ -0,0 +1,50 @@
...
+        os: [ubuntu-latest]
+        python-version: ["3.10"]
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        id: setup_python
+        with:
+          python-version: ${{ matrix.python-version}}
```
KennethEnevoldsen, line 26:
```suggestion
    steps:
      - uses: actions/checkout@v3

      - name: Set up Python
        uses: actions/setup-python@v4
        with:
          python-version: "3.10"
```

simplifying a bit more here. No need for the test grid.

.github/workflows/snyk.yml
```diff
@@ -0,0 +1,50 @@
...
+        shell: bash
+        run: |
+          pip install invoke
+          pip install -r requirements.txt -r test-requirements.txt -r dev-requirements.txt -r gpu-requirements.txt
+
+      - name: Install snyk
+        run: npm install -g snyk
+
+      - name: Run Snyk to check for vulnerabilities
+        id: snyk
```
KennethEnevoldsen, line 38:
```suggestion
```

KennethEnevoldsen (commented):
I jumped in a did a few simplifications on the ci. Looks much simpler without the comment. Glad we made that decision.

.github/workflows/snyk.yml
sarakolding: i'm good with the name change

.github/workflows/snyk.yml
sarakolding: yup, it needs authentication

tasks.py
```diff
@@ -524,6 +524,9 @@ def snyk(c: Context):
         "gpu-requirements.txt",
         "test-requirements.txt",
     ]:
+        if Path(requirements_file).exists() is False:
+            raise FileNotFoundError("One of the provided files could not be found.")
```
MartinBernstorff, line 528:
```suggestion
            raise FileNotFoundError(f"{requirements_file} does not exist")
```

MartinBernstorff:
> Lasse and I agreed that we would keep the hardcoded file paths, to keep things consistent. 

In that case, we should only have it hard-coded in one place. Or, alternatively, infer them all with `rglob`.

Makes it much easier to maintain 👍 

The snyk workflow is also failing now. Other than that, looks good to me. Good work!

.github/workflows/snyk.yml
```diff
@@ -0,0 +1,42 @@
...
+          pip install invoke
+          pip install -r requirements.txt -r test-requirements.txt -r dev-requirements.txt -r gpu-requirements.txt
+
+      - name: Install snyk
+        run: npm install -g snyk
+
+      - name: Run Snyk to check for vulnerabilities
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+        continue-on-error: false
```
KennethEnevoldsen, line 33:
```suggestion
```

sarakolding:
have added some globbing to install the requirement files, but it's not super sleek. suggestions or good enough?

tasks.py
```diff
@@ -517,22 +517,27 @@ def pr(c: Context, auto_fix: bool = True, create_pr: bool = True):
...
-        "test-requirements.txt",
-    ]:
-        if Path(requirements_file).exists() is False:
-            raise FileNotFoundError(f"{requirements_file} does not exist")
+def vulnerability_scan(c: Context):
+    requirements_files = Path().parent.rglob("*requirements.txt")
+    top_level_files = [file for file in requirements_files if file.parent == Path()]
+    for requirements_file in top_level_files:
+        if requirements_file.exists() is False:
+            raise FileNotFoundError(f"{requirements_file} does not exist.")
```
MartinBernstorff, line 525: Is there a need to only get top-level files? And if so, we should just `glob`, not `rglob` 👍 

```python
requirements_files = list(Path(__file__).rglob("*requirements.txt"))
for req in requirements_files:
    ...
```

MartinBernstorff: No need to check the file exists anymore either; it won't be found by glob if it doesn't exist.

tasks.py
```diff
@@ -517,22 +517,27 @@ def pr(c: Context, auto_fix: bool = True, create_pr: bool = True):
...
 
         c.run(
             f"snyk test --file={requirements_file} --package-manager=pip",
             pty=NOT_WINDOWS,
         )
 
 
+@task
+def install_requirements(c: Context):
+    requirements_files = Path().parent.rglob("*requirements.txt")
```
MartinBernstorff, line 535: Same as above.

.github/workflows/static_type_checks.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         shell: bash
         run: |
           pip install invoke
-          pip install -r requirements.txt -r test-requirements.txt -r dev-requirements.txt -r gpu-requirements.txt
+          inv install-requirements
```
KennethEnevoldsen, line 33: God idé!

KennethEnevoldsen (commented):
Ser virkelig godt ud!

MartinBernstorff:
Mega godt arbejde Sara! Super glad for resultatet 👍

PR #492, opened 2023-11-27
feat: create split based on region

HLasse:
# Notes for reviewers
Fixes #314

PR #496, opened 2023-11-29
feat: create split based on id

### LINKED ISSUE #494: feat: make split based on ids
Make a `DataFilter` for splitting based on the original ids

HLasse:
Fixes #494 

Depends on #492 so might want to take a look there 👍


psycop/common/model_training_v2/trainer/data/data_filters/original_ids.py
```diff
@@ -0,0 +1,66 @@
...
+            match split_name:
+                case "train":
+                    split_names.append(SplitName.TRAIN)
+                case "val":
+                    split_names.append(SplitName.VALIDATION)
+                case "test":
+                    split_names.append(SplitName.TEST)
+                case _:  # pyright: ignore [reportUnnecessaryComparison]
+                    raise ValueError(
+                        f"Splitname {split_name} is not allowed, try from ['train', 'test', 'val']",
```
KennethEnevoldsen, line 62: Is that syntax for "otherwise"? (that is a really bad syntax - can we do something else?)

psycop/common/model_training_v2/trainer/data/data_filters/id.py
KennethEnevoldsen, line 1: delete?

KennethEnevoldsen (approved):
Some more info on the PR would be great (you can use copilot for it in vscode). The issue doesn't give a lot of info.

psycop/common/model_training_v2/trainer/data/data_filters/test_data_filters.py
```diff
@@ -37,3 +40,15 @@ def test_regional_filter():
     filtered_df = regional_filter.apply(dataloader).collect()
 
     assert filtered_df.shape == (3, 2)
+
+
+def test_id_filter():
+    split_sequence = pl.Series("dw_ek_borger", [1])
+    dataloader = MockDataLoaderForFilters()
```
KennethEnevoldsen, line 47: Why not just call this a df? instead of a dataloader? isn't the intention with lazy object that it should be treated as the object (a dataframe).

HLasse:
Fair point re more info! The idea is basically that we want it to be easy to switch between different train/val/test splits, e.g. based on the original ID split that we made a while ago (this PR) or by geography as we talked about in a recent group meeting #492 , by calendar time, etc.

These PRs introduce the FilteredDataloader that achieves this. You simply load your entire dataset (generated with e.g. timeseriesflattener) using a dataloader and use your desired `DataFilter` to subset the datasets. This makes it super easy to test across the different types of splits we have

psycop/common/model_training_v2/trainer/data/data_filters/original_ids.py
HLasse: Yep, that's the syntax. Could do a bunch of ifs but don't feel that's much better

psycop/common/model_training_v2/trainer/data/data_filters/test_data_filters.py
HLasse: Dataloaders are more flexible in that we can for instance use the `ParquetVerticalConcatenator` to load from multiple files (e.g. splits), or a horizontal concatenater to combine different feature sets.

PR #505, opened 2023-12-01
dev: refactor taskspy into maintainable modules

MartinBernstorff:
@KennethEnevoldsen This should make maintenance (and potentially future simplification) much easier.

I'd love feedback on:
* The chosen structure

And help on:
* Pylance reports unresolved imports within the automation modules, in spite of them having an `__init__.py` and using relative imports. The tasks still run fine. Do you have suggestions here?

### DIFF
--- a/automation/logger.py
+++ b/automation/logger.py
@@ -7,8 +7,7 @@ from typing import Optional
 
 from invoke import Context, Result, task
 
-from automation.git import is_uncommitted_changes
-from automation.windows import NOT_WINDOWS
+from .windows import NOT_WINDOWS
 
 
 class MsgType:
--- a/automation/logger.py
+++ b/automation/logger.py
@@ -1,3 +1,16 @@
+import multiprocessing
+import platform
+import re
+import shutil
+from pathlib import Path
+from typing import Optional
+
+from invoke import Context, Result, task
+
+from automation.git import is_uncommitted_changes
+from automation.windows import NOT_WINDOWS
+
+
 class MsgType:
     # Emojis have to be encoded as bytes to not break the terminal on Windows
     @property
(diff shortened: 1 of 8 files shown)


MartinBernstorff:
@KennethEnevoldsen, just a friendly reminder here :-)

MartinBernstorff:
@KennethEnevoldsen I'd also love some comments on the top-level comment, specifically:

And help on:
* Pylance reports unresolved imports within the automation modules, in spite of them having an __init__.py and using relative imports. The tasks still run fine. Do you have suggestions here?

Future suggestions:
* Do we need the functionality in the setup.py? Seems that for the "fully automatic" usecase, we should use Docker. For cases where docker doesn't fit, we could make do with just the 'install' entrypoint.
* Pre-commit adds some complexity in lint.py and keeping versions aligned in pyproject.toml and pre-commit-config.yaml. Perhaps we should leave it behind, and just use the native packages?

automation/logger.py
KennethEnevoldsen, line 1: move automation to `src/automation`?

MartinBernstorff: Yeah, that's a good quick fix!

It seems wrong; it's not really a part of the package. But doesn't seem worth it to dig into right now.

MartinBernstorff:
> In the tasks.py file? What is the error?

Moving the `automation` module into the `psycop` package solved it. It's a bit of a hack, but I think it will do for now :-)

> Hmm right, yea we might have outgrown pre-commit. We can simply switch to inv lint. It does make CI very manageable, but it does (as you say) lead to inconveniences in versioning. The exception would be if anyone wants to keep their pre-commit hooks?

Agreed. After sleeping on it, I don't this is important now, but perhaps the next time we get into trouble with it :-)

automation/logger.py
KennethEnevoldsen: Ahh yes we don't have a src folder. That makes it a bit weirder I agree.

PR #509, opened 2023-12-01
feat: add disk logger

### LINKED ISSUE #508: feat: implement a disk logger
We often want to log the cfg used for a model to disk for e.g. evaluation. A disk logger (which would also save stdout/err) would be needed

HLasse:
Add disk logger. Fixes #508

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -6,7 +6,7 @@
         {
             "label": "PR",
             "type": "shell",
-            "command": "make push && inv create-pr",
+            "command": "git push && inv create-pr",
             "group": "test",
             "dependsOn": [
                 "Lint",
@@ -24,7 +24,7 @@
         {
             "label": "Merge main",
             "type": "shell",
-            "command": "make merge-main && make push",
+            "command": "inv merge-main && git push",
             "presentation": {
                 "reveal": "always",
                 "group": "pr",
--- a/psycop/automation/environment.py
+++ b/psycop/automation/environment.py
@@ -1,0 +1,14 @@
+import platform
+
+NOT_WINDOWS = platform.system() != "Windows"
+
+
+def on_ovartaci() -> bool:
+    import platform
+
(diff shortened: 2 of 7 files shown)


psycop/common/model_training_v2/loggers/disk_logger.py
```diff
@@ -0,0 +1,58 @@
...
+from psycop.common.model_training_v2.loggers.base_logger import BaselineLogger
+from psycop.common.model_training_v2.trainer.task.base_metric import CalculatedMetric
+
+
+@BaselineRegistry.loggers.register("disk_logger")
+class DiskLogger(BaselineLogger):
+    def __init__(self, experiment_path: str) -> None:
+        self.experiment_path = Path(experiment_path)
+        self.log_path = self.experiment_path / "logs.log"
+        self.cfg_log_path = self.experiment_path / "config.json"
```
MartinBernstorff, line 16: Hardcode these where they are used? Improves cohesion :-)

psycop/common/model_training_v2/loggers/disk_logger.py
```diff
@@ -0,0 +1,58 @@
...
+            "%(asctime)s %(levelname)s %(name)s %(message)s",
+            datefmt="%Y-%m-%d %H:%M:%S",
+        )
+        fh.setFormatter(formatter)
+
+        # Add the handler to the logger
+        logger.addHandler(fh)
+
+        logger.info("Disk logger initialized")
+        return logger
```
MartinBernstorff, line 58: The logging module is so complicated, so bear with me:

As I understand, the logger is now in the namespace "GLOBAL.DiskLogger". Any new logger from logging that handles the GLOBAL namespace will also get log messages from this logger.

It don't think that's a problem, but just wanted to point to the complexity 👍

MartinBernstorff (approved):
Looks great Lasse! _Tiny_ considerations, looks ready to merge 👍

PR #519, opened 2023-12-05
ci: only run vulnerability scan only when requirements are modified

sarakolding:
# Notes for reviewers


sarakolding:
@MartinBernstorff i get an error that i cannot google my way out of 🥲

psycop/automation/git.py
```diff
@@ -18,13 +19,13 @@ def is_uncommitted_changes(c: Context) -> bool:
     return uncommitted_changes
 
 
-def filetype_modified_since_head(c: Context, file_suffix: str) -> bool:
+def filetype_modified_since_main(c: Context, regex_pattern: re.Pattern[str]) -> bool:
```
MartinBernstorff, line 22: Probably want this to take a str, and then it compiles within the function 👍

tasks.py
```diff
@@ -157,13 +161,27 @@ def automerge(c: Context):
...
+            print(
+                "🟢 No requirements.txt files modified since main, skipping vulnerability scan",
+            )
+
+    else:
+        for requirements_file in requirements_files:
+            c.run(
+                f"snyk test --file={requirements_file} --package-manager=pip",
+                pty=NOT_WINDOWS,
+            )
```
MartinBernstorff, line 184: Can be implemented as a guard, decreasing conditionality and removing the else-statement:

```suggestion
        if not filetype_modified_since_main(c, re.compile(r"requirements\.txt$")):
            print(
                "🟢 No requirements.txt files modified since main, skipping vulnerability scan",
            )
            return None
            
        for requirements_file in requirements_files:
            c.run(
                f"snyk test --file={requirements_file} --package-manager=pip",
                pty=NOT_WINDOWS,
            )
```

MartinBernstorff:
I've added a few comments to the code.

The error was caused by the GitHub Actions Runner only checking out the PR branch, so `main` was not present on the runner. We now explicitly fetch `main` so we can compare against it 👍

PR #520, opened 2023-12-05
feat: artifact saver

### LINKED ISSUE #510: Add 'on_training_end' to loggers
We would like to be flexible on what we do post-training - e.g. Just logging the metric, or saving the pipe to disk.

One option could be to add a required 'on_training_end' function to the loggers that would be called after training and could be configured based on the individual loggers and the needs of the user. 

Could also add a PostTraining class that handles this, or something completely different. Very open for ideas

HLasse:
Adds `artifact_savers` to registry and the base class `BaselineArtifactSaver` along with an implementation of `TaskDiskSaver` that saves the task to disk. 

Fixes #510


psycop/common/model_training_v2/trainer/split_trainer.py
```diff
@@ -82,11 +87,22 @@ def train(self) -> TrainingResult:
...
+
+        return training_result
+
+    def _save_artifacts(self, training_result: TrainingResult) -> None:
+        if self.artifact_savers is not None:
+            for artifact_saver in self.artifact_savers:
+                artifact_saver.save_artifact(
+                    trainer=self,
+                    training_result=training_result,
+                )
```
MartinBernstorff, line 108: Might want to log a warning here if no artifact-savers are specified 👍

psycop/common/model_training_v2/artifact_savers/task_disk_saver.py
```diff
@@ -0,0 +1,28 @@
...
+from psycop.common.model_training_v2.config.baseline_registry import BaselineRegistry
+from psycop.common.model_training_v2.trainer.base_trainer import (
+    BaselineTrainer,
+    TrainingResult,
+)
+
+
+@BaselineRegistry.artifact_savers.register("task_disk_saver")
+class TaskDiskSaver(BaselineArtifactSaver):
+    def __init__(self, experiment_path: str):
```
MartinBernstorff, line 16: Perhaps this should signify which artifacts it is capable of saving, e.g. using arguments to init? E.g. `save_trainer`, `save_metric` etc.? Is there anything else in TrainingResult we might save?

psycop/common/model_training_v2/artifact_savers/test_task_disk_saver.py
```diff
@@ -0,0 +1,65 @@
...
+        task=BinaryClassificationTask(
+            pred_time_uuid_col_name="pred_time_uuid",
+            task_pipe=BinaryClassificationPipeline(
+                sklearn_pipe=Pipeline([logistic_regression_step()]),
+            ),
+        ),
+        logger=TerminalLogger(),
+        metric=BinaryAUROC(),
+    )
+    training_result = TrainingResult(
```
MartinBernstorff, line 54: Convert to integration test

MartinBernstorff (commented):
General minor comments from previously; might be solved if we look at how we want to evaluate.

MartinBernstorff:
This might be relevant for #523, but closing this until we are ready to work on it. (The mention means it is automatically reference in #523).

Feel free to re-open if you disagree 👍

PR #528, opened 2023-12-11
docs: minor renames in behrt

MartinBernstorff:
1. Rename `filter_events` to `A_diagnoses_only` to convey content. This makes it less general; if we want to avoid that, perhaps we should conform to a `Protocol`?
2. Add documentation of Caliber codes


KennethEnevoldsen (approved):
Looks fine. Filter and reformat event is the "general" part. I believe. We should probably create a protocol regardless.

PR #530, opened 2023-12-11
feat: finetune sequence models for t2d

MartinBernstorff:


psycop/common/sequence_models/tasks/tasks.py
```diff
@@ -77,8 +79,8 @@ def forward(  # type: ignore
         inputs: dict[str, torch.Tensor],
         masked_lm_labels: torch.Tensor,
     ) -> dict[str, torch.Tensor]:
-        embedded_patients = self.embedding_module(inputs)
-        encoded_patients = self.encoder_module(
+        embedded_patients = self.embedder(inputs)
+        encoded_patients = self.encoder(
```
MartinBernstorff, line 83: Here we run the `__call__` method on the encoder, right? That method must take a src and  src_key_padding_mask. Does that imply more requirements than it being an `nn.Module`?

PR #563, opened 2023-12-15
feat: pretrain with `mlflow`

### LINKED ISSUE #562: feat: unify logger interfaces
Makes it super easy to switch between loggers; just change the registry key.

MartinBernstorff:
For #560, I needed the path of the pretrained model. However, I think it is on your user account (ADMINKEN**) @KennethEnevoldsen.

So, wanted to run a new pretraining that saves to a dir on `E:/shared_resources`, and might as well add mlflow at the same time.

* Unified logger interfaces, to make it easier to change between them
* Added mlflow logger

Started a training, but failed because of the MLflow server being down. More info here: https://github.com/Aarhus-Psychiatry-Research/psycop-ml-ops/issues/54

Also fixes #562.

### DIFF
--- a/psycop/common/sequence_models/logger.py
+++ b/psycop/common/sequence_models/logger.py
@@ -23,67 +23,36 @@ def handle_wandb_folder():
         )
 
 
-@runtime_checkable
-class LoggerFactory(Protocol):
-    def __init__(
-        self,
-        save_dir: Path | str,
-        experiment_name: str,
-        run_name: Optional[str] = None,
-        offline: bool = False,
-    ) -> None:
-        ...
-
-    def get_logger(
-        self,
-    ) -> plLogger:
-        ...
-
-
 @Registry.loggers.register("wandb")
-class WandbCreator(LoggerFactory):
-    def __init__(
-        self,
-        save_dir: Path | str,
-        experiment_name: str,
-        offline: bool,
-        run_name: Optional[str] = None,
-    ) -> None:
-        handle_wandb_folder()
+def create_wandb_logger(
+    save_dir: Path | str,
+    experiment_name: str,
+    offline: bool,
+    run_name: Optional[str] = None,
+) -> plLogger:
+    handle_wandb_folder()
 
-        self.save_dir = save_dir
-        self.experiment_name = experiment_name
-        self.run_name = run_name
-        self.offline = offline
-
-    def get_logger(self) -> plLogger:
-        return plWandbLogger(
-            name=self.run_name,
-            save_dir=f"{self.save_dir}/{self.experiment_name}/{self.run_name}",
-            offline=self.offline,
-            project=self.experiment_name,
-        )
+    return plWandbLogger(
+        name=run_name,
+        save_dir=f"{save_dir}/{experiment_name}/{run_name}",
+        offline=offline,
+        project=experiment_name,
+    )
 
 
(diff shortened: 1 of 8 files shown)


psycop/common/sequence_models/logger.py
```diff
@@ -18,30 +23,67 @@ def handle_wandb_folder():
...
-        version=version,
-        offline=offline,
-        dir=dir,
-        id=id,
-        anonymous=anonymous,
-        project=project,
-        prefix=prefix,
-        checkpoint_name=checkpoint_name,
-    )
+class WandbCreator(LoggerFactory):
```
KennethEnevoldsen, line 44: Any reason why we don't simpy just return the logger object?

KennethEnevoldsen: that way we don't assume the `.get_logger` method

psycop/common/sequence_models/train.py
```diff
@@ -47,27 +47,32 @@ def train(config_path: Path | None = None) -> None:
...
     config = parse_config(config_dict)
 
+    # Setup the logger and pass it to the TrainingConfig
     training_cfg = config.training
+
+    logger_factory = config.logger_factory
+    if logger_factory is not None:
+        logger = config.logger_factory.get_logger()  # type: ignore
+        training_cfg.trainer.logger = logger
+        training_cfg.trainer.Config.allow_mutation = False
```
KennethEnevoldsen, line 57: Seems like it might cause an issue? (see comment above)

KennethEnevoldsen (changes requested):
Let us discuss the factory - there is probably a good reason for it that I am missing.

psycop/common/sequence_models/logger.py
MartinBernstorff: The idea was to unify the `__init__` interfaces, so that swapping from `wandb` to `mlflow` was just a case of swapping which object in the registry you are pointing towards. This is much easier using a protocol than a Callable type-hint.

OTOH, it's probably a relatively rare use-case, so I'm up for either.

psycop/common/sequence_models/logger.py
```diff
@@ -20,28 +22,34 @@ def handle_wandb_folder():
...
-    anonymous: Optional[bool] = None,
-    project: Optional[str] = None,
-    prefix: str = "",
-    checkpoint_name: Optional[str] = None,
-) -> WandbLogger:
+    save_dir: Path | str,
+    experiment_name: str,
+    offline: bool,
+    run_name: Optional[str] = None,
+) -> plLogger:
```
KennethEnevoldsen, line 29:
```suggestion
) -> WandbLogger:
```

psycop/common/sequence_models/logger.py
```diff
@@ -20,28 +22,34 @@ def handle_wandb_folder():
...
+    )
+
+
+@Registry.loggers.register("mlflow")
+def create_mlflow_logger(
+    save_dir: Path | str,
+    experiment_name: str,
+    offline: bool = False,
+    run_name: Optional[str] = None,
+) -> plLogger:
```
KennethEnevoldsen, line 46:
```suggestion
) -> plMLFlowLogger:
```

KennethEnevoldsen (approved):
Looks good

PR #566, opened 2023-12-15
feat: convert `patient slice getters` to classes

MartinBernstorff:
Depends on #563, review that first.

Fixes #561.

* Shared protocol to make composition much easier
* Separates loading and init, to make for faster config validation

PR #574, opened 2023-12-15
feat: first mlflow pretraining

MartinBernstorff:
Depends on #566, review that first.

It works! 

image


KennethEnevoldsen (commented):
hard to see differences here from the other. Unsure what is different.

MartinBernstorff:
My pseudo-stacking did not work out at all. Had to combine the branches, this one should be reviewable now.

KennethEnevoldsen (approved):
looks good

PR #576, opened 2023-12-18
fix critical vulnerability in mlflow

sarakolding:
Vulnerability to directory traversal found in certain versions of mlflow (https://security.snyk.io/vuln/SNYK-PYTHON-MLFLOW-6128849). Can be fixed by upgrading to >=2.9.2.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ hydra-joblib-launcher==1.2.0
 hydra-optuna-sweeper==1.2.0
 ipython==7.34.0
 Levenshtein==0.21.0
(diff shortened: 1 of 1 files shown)


MartinBernstorff:
Fixing with an upgrade is definitely the way to go here.

Note also that this is a vulnerability in the MLflow server, which we do not run. This means it is not important for us. In cases where there is not an upgrade available which fixes it, we would typically get an exemption from Søren+Andreas 👍

sarakolding:
You're welcome, Martinh ;-)

PR #602, opened 2023-12-22
faster generation using slots

KennethEnevoldsen:
This should help creating instances faster in behrtembedder.reformat. Dataclass creation is by far the most time consuming thing. Second is mapicd which takes up 16% of the time.

### DIFF
--- a/psycop/common/data_structures/patient.py
+++ b/psycop/common/data_structures/patient.py
@@ -129,7 +129,7 @@ class Patient:
         return prediction_sequences
 
 
-@dataclass(frozen=True)
+@dataclass(frozen=True, slots=True)
 class PatientSlice:
     patient: Patient
     temporal_events: Sequence[TemporalEvent]
--- a/psycop/common/data_structures/temporal_event.py
+++ b/psycop/common/data_structures/temporal_event.py
@@ -7,7 +7,7 @@ if TYPE_CHECKING:
     import datetime as dt
 
 
-@dataclass
+@dataclass(slots=True)
 class TemporalEvent:
     """
     Attributes:
--- a/psycop/common/sequence_models/embedders/BEHRT_embedders.py
+++ b/psycop/common/sequence_models/embedders/BEHRT_embedders.py
@@ -245,7 +245,7 @@ class BEHRTEmbedder(nn.Module, PatientSliceEmbedder):
                     value=mapped_value,
                 )
                 for e in filtered_events
-                if (mapped_value := self.map_icd10_to_caliber(e.value)) is not None
+                if (mapped_value := self.map_icd10_to_caliber(e.value)) is not None  # type: ignore
             ]
 
             _patient_slices.append(PatientSlice(p.patient, temporal_events))
--- a/psycop/common/sequence_models/embedders/BEHRT_embedders.py
+++ b/psycop/common/sequence_models/embedders/BEHRT_embedders.py
@@ -7,6 +7,7 @@ import logging
 from collections.abc import Sequence
 from dataclasses import dataclass, field
 from datetime import datetime
+from functools import lru_cache
 
 import numpy as np
 import torch
@@ -208,6 +209,7 @@ class BEHRTEmbedder(nn.Module, PatientSliceEmbedder):
     def is_A_diagnosis(self, event: TemporalEvent) -> bool:
         return event.source_type == "diagnosis" and event.source_subtype == "A"
 
+    @lru_cache
     def map_icd10_to_caliber(
         self,
         diagnosis_code: str,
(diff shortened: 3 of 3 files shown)


KennethEnevoldsen:
the changes I added:

- **1 min. 15 seconds (no-changes)**
- 39 seconds (only slots) <-- so def. the data classes that was the problem.
- **33.2 seconds (after changes all changes)**
- 37.4 seconds (without LRU cache)

I consider removing lru_cache, but I am unsure


I tested using the code:

```
import datetime as dt

from psycop.common.data_structures import TemporalEvent
from psycop.common.data_structures.patient import Patient
from psycop.common.sequence_models.embedders.BEHRT_embedders import BEHRTEmbedder

embedding_module = BEHRTEmbedder(d_model=384, dropout_prob=0.1, max_sequence_length=128)
# Check that diagnosis codes that are not in the mapping are excluded
# (this patient has no diagnosis codes in the mapping)
patient = Patient(
    patient_id=11,
    date_of_birth=dt.datetime(year=1990, month=1, day=1),
)

# Add temporal events to check that diagnosis codes are mapped correctly
temporal_events = [
    TemporalEvent(
        timestamp=dt.datetime(2021, 1, 1),
        value="A00",
        source_type="diagnosis",
        source_subtype="A",
    ),
    # Check that two different ICD10 codes map to the same caliber code (A00, A30 -> Bacterial Diseases (excl TB))
    TemporalEvent(
        timestamp=dt.datetime(2021, 1, 3),
        value="A30",
        source_type="diagnosis",
        source_subtype="A",
    ),
    # Check that diagnoses with different subtype than A are excluded
    TemporalEvent(
        timestamp=dt.datetime(2021, 1, 3),
        value="A30",
        source_type="diagnosis",
        source_subtype="B",
    ),
    # Check that ICD10 code maps to different caliber code than those above (I65 -> Transient ischaemic attack)
    TemporalEvent(
        timestamp=dt.datetime(2021, 1, 3),
        value="I65",
        source_type="diagnosis",
        source_subtype="A",
    ),
    # Check that F909 is mapped to F90 (Hyperkinetic disorders) since F909 is not in the mapping
    TemporalEvent(
        timestamp=dt.datetime(2021, 1, 3),
        value="F909",
        source_type="diagnosis",
        source_subtype="A",
    ),
]

patient.add_events(temporal_events * 20)

patient_slices_mapped = embedding_module.reformat(
    [patient.as_slice()] * 500_000,
)
```

MartinBernstorff:
Looks great! A handful of ramblings:

* It's quite surprising to me that `slots` alone hanged performance that much! According to [this post](https://stackoverflow.com/questions/55248416/python-3-which-one-is-faster-for-accessing-data-dataclasses-or-dictionaries), it should only improve dataclass initialisation performance by ~15%. 

* I wonder whether using `[patient] * 500_000` does not result in 500_000 patients, but 500_000 references/pointers to the same patient. This means very little/no time is spent moving memory around on the heap, so all the time is spent initialising dataclasses. That would make any change to dataclass-init performance dwarf other performance changes. 

* We could probably get rid of the `B019` by making it a static method.

* Because the examples only have 5 different codes, I wonder whether the system itself is doing some efficient caching on the mapping. LRU cache's default size seems to be 128: https://docs.python.org/3/library/functools.html, so if we're going to use it on real data, we want to bump that.

Looking forward to benchmarking it on some real data in the new year, or just merging this in and calling it a day! 🎉

KennethEnevoldsen:
> We could probably get rid of the B019 by making it a static method.

yea it was just that the gains from LRU caching were minor (even in a case where we were certain to have hits)

> Because the examples only have 5 different codes, I wonder whether the system itself is doing some efficient caching on the mapping. LRU cache's default size seems to be 128: https://docs.python.org/3/library/functools.html, so if we're going to use it on real data, we want to bump that.

I don't think adding a lot more than 128 is gonna be nec. - the primary gain is in the most frequent, I imagine 128 covers a wide range.

Btw. another option here is to precompute all variations of the dict (so there was no reason to do multiple-lookups (you just know that for this code it will be reduced and then return X so you might as well just return X). This is essentially what the LRU cache would do.


> I wonder whether using [patient] * 500_000 does not result in 500_000 patients, but 500_000 references/pointers to the same patient. This means very little/no time is spent moving memory around on the heap, so all the time is spent initializing dataclasses. That would make any change to dataclass-init performance dwarf other performance changes.

I have added converted to:

```
patients = [create_patient(i) for i in range(500_000)]
```

Where create_patients wraps all the creation in a function and uses I as the patient id.

Time with changes: 47s 
Time without changes: 1m 39,5

Still seems like slots are pretty great.

MartinBernstorff:
Ah yeah, so should we benchmark on real data before merging?

MartinBernstorff:
Again, super happy you've worked on this 👍

MartinBernstorff:
Ah, lo, you've assigned me. I assume you'd li,e me to test it out?

PR #611, opened 2024-01-02
fix: replace disallowed symbols in config before logging to mlflow

HLasse:
Mlflow does not allow @ and * when logging params. Changed the values to "" and "_" instead. Added the name of the run-id to tmp path to avoid potential conflicts during parallel runs. 

Merged, but feel free to post review


MartinBernstorff:
Huh, I distinctly remember fixing this because I encountered the same error, but cannot find it in any merged PRs. Perhaps I have it on an unmerged branch.

Your fix looks great! Nice idea doing it after flattening, means you don't need recursion in the replacement function 👍

Issue #628, opened 2024-01-09
feat: SplitLoaders should return SplitFrames


MartinBernstorff:
* Write the code
* Validate on Ovartaci

PR #632, opened 2024-01-09
feat: SplitLoaders should return SplitFrames (issue #628)

MartinBernstorff:


MartinBernstorff:
Current dependencies on/for this PR:
* `main`
  * **PR #632** <a href="https://app.graphite.dev/github/pr/Aarhus-Psychiatry-Research/psycop-common/632?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/Aarhus-Psychiatry-Research/psycop-common/632?utm_source=stack-comment).

HLasse (commented):
LGTM. The question of moves outside the region still haunts me a bit, as using `load_stratified_by_region_split_ids` naively (without filtering prediction times) could lead to leakage

MartinBernstorff:
Strong agree; this probably also indicates that we should make it clearer which split-functions are public and which are private. E.g. we never want "users" to naively use the geographical split IDs. 

Perhaps it should be contained in the class that acts as a filter. Let's look at it.

PR #637, opened 2024-01-10
fix: registry testing project-specific registered functions

HLasse:
The test for whether valid configs exist and whether all registered functions can be resolved now only checks functions/classes that are in "common"

### DIFF
--- a/psycop/projects/scz_bp/model_training/estimator_steps/imputers.py
+++ b/psycop/projects/scz_bp/model_training/estimator_steps/imputers.py
@@ -3,10 +3,9 @@ from typing import Literal
 from sklearn.ensemble import RandomForestRegressor
 from sklearn.experimental import enable_iterative_imputer  # noqa
 from sklearn.impute import IterativeImputer, SimpleImputer
-from psycop.common.model_training_v2.config.baseline_registry import BaselineRegistry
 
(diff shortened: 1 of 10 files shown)


psycop/projects/scz_bp/model_training/estimator_steps/imputers.py
```diff
@@ -1,6 +1,7 @@
 from typing import Literal
 
 from sklearn.ensemble import RandomForestRegressor
+from sklearn.experimental import enable_iterative_imputer  # noqa
```
MartinBernstorff, line 4: Why a new edit here?

MartinBernstorff (approved):
So satisfying when code you've written turns out to be easily extendable! ❤️

One comment, then LGTM!

psycop/projects/scz_bp/model_training/estimator_steps/imputers.py
HLasse: The `IterativeImputer` (used for missforest) is "experimental" which means that one needs to explicitly do the import above for it to work (import `IterativeImputer` is apparently not enough). Got an error when testing the `populate_scz_bp registry` function, but should have been in a different PR - sorry.

Issue #699, opened 2024-01-16
feat: optuna hparam optimization


HLasse:
We want to be able to run (parallel) training of multiple multiple models for hyperparameter optimization. 

**Use cases**
- Specify a space to search over and get the scores + the best config out 

**Implementation**
- Start an optuna study with user-specified name
- Take a `ConfigSpace`: a `Config` with `SuggesterSpace`s somewhere
- Suggest hyperparameters from each space to create an instance of a `ConcreteConfig`
- Create random name for mlflow-run (if optuna creates one for their trial, use it)
- Add mlflow logger to the `ConcreteConfig` with the generated name
- Run training 
- Save config to disk and log results

PR #712, opened 2024-01-16
feat(#699): optuna-hparam-optimization

HLasse:

### DIFF
--- a/psycop/common/model_training_v2/hyperparameter_suggester/optuna_wrapper.py
+++ b/psycop/common/model_training_v2/hyperparameter_suggester/optuna_wrapper.py
@@ -1,0 +1,8 @@
+
+
+
+
+from typing import Literal
+
+
+def create_optuna_study(name: str, direction: Literal["minimize", "maximize"] = "maximize", pruner)


HLasse:
Current dependencies on/for this PR:
* `main`
  * **PR #712** <a href="https://app.graphite.dev/github/pr/Aarhus-Psychiatry-Research/psycop-common/712?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/Aarhus-Psychiatry-Research/psycop-common/712?utm_source=stack-comment).

psycop/common/model_training_v2/hyperparameter_suggester/test_optuna_hyperparameter_search.py
```diff
@@ -0,0 +1,93 @@
...
+                l1_ratio={"low": 0.1, "high": 1, "logarithmic": False},
+                solvers=("saga", "lbfgs"),
+            ),
+        },
+    }
+    assert (
+        OptunaHyperParameterOptimization()._validate_configspace(  # type: ignore
+            cfg,
+        )
+        is None
```
MartinBernstorff, line 60: Should this return anything? (Maybe we don't need the assertion, just a "not pytest.raises" or something?)

psycop/common/model_training_v2/hyperparameter_suggester/test_optuna_hyperparameter_search.cfg
```diff
@@ -0,0 +1,48 @@
...
+
+[trainer.task.task_pipe.sklearn_pipe.*.logistic_regression]
+@suggesters = "logistic_regression_suggester"
+solvers = ["saga"]
+penalties = ["l1","l2","elasticnet"]
+
+[trainer.task.task_pipe.sklearn_pipe.*.logistic_regression.C]
+low = 0.00000001
+high = 10
+logarithmic = false
```
MartinBernstorff, line 43: After reading this, perhaps we should make the parameters a 3-item list? This would allow us to specify:

`C = [0.000001, 10, false]` 

Instead of a full section. Not necessarily in this PR, though 👍

psycop/common/model_training_v2/hyperparameter_suggester/optuna_hyperparameter_search.py
```diff
@@ -0,0 +1,108 @@
...
+
+        run_result = concrete_config_schema.trainer.train()
+        return run_result.metric.value
+
+    @staticmethod
+    def conduct_hyperparameter_optimization_from_file(
+        cfg_file: Path,
+        n_trials: int,
+        n_jobs: int,
+        direction: Literal["maximize", "minimize"] = "maximize",
```
MartinBernstorff, line 88: I think having a default here could be dangerous. Perhaps it should even be specified in the ConfigSpace .cfg?

psycop/common/model_training_v2/hyperparameter_suggester/optuna_hyperparameter_search.py
```diff
@@ -0,0 +1,108 @@
...
+    suggest_hyperparams_from_cfg,
+)
+from psycop.common.model_training_v2.hyperparameter_suggester.suggesters.base_suggester import (
+    Suggester,
+)
+
+
+class OptunaHyperParameterOptimization:
+    @staticmethod
+    def _validate_configspace(cfg: dict[str, Any]) -> None:
```
MartinBernstorff, line 22: Perhaps rename to make clear that the only validation it does is to check that it has suggesters?

psycop/common/model_training_v2/hyperparameter_suggester/optuna_hyperparameter_search.py
```diff
@@ -0,0 +1,108 @@
...
+            trial=trial,
+        )
+        concrete_config_schema = BaselineSchema(
+            **BaselineRegistry.resolve(concrete_config),
+        )
+
+        concrete_config_schema.logger.log_config(Config(concrete_config))
+
+        run_result = concrete_config_schema.trainer.train()
+        return run_result.metric.value
```
MartinBernstorff, line 81: The value of dataclasses ❤️

psycop/common/model_training_v2/hyperparameter_suggester/optuna_hyperparameter_search.py
```diff
@@ -0,0 +1,108 @@
...
+
+    @staticmethod
+    def _resolve_only_suggestors(cfg: dict[str, Any]) -> dict[str, Any]:
+        """Resolves only suggesters in a nested config. Suggesters are identified
+        by the presence of the @suggestors key."""
+        cfg_copy = copy.deepcopy(cfg)
+
+        for key, value in cfg_copy.items():
+            if isinstance(value, dict):
+                if "@suggesters" in value:
```
MartinBernstorff, line 55: Might want to expose this on either the method or self? Ideally taking a list[str] so we can co-locate the suggesters and the implementations (e.g. `estimater_step_suggesters` etc.).

Or we could have it take a regex, like `*._suggesters$`.

In case we:
* Want to reuse this somewhere else, or
* Have to change it in the future

MartinBernstorff (commented):
This looks really good, Lasse! 

Good thing we didn't underestimate the amount of work needed for glue-code like this. I hope the design-session made this easier 👍 

A handful of comments, let me know what you think.

psycop/common/model_training_v2/hyperparameter_suggester/test_optuna_hyperparameter_search.py
HLasse: Agreed, removed the assertion. The test will fail if an error is raised

psycop/common/model_training_v2/hyperparameter_suggester/test_optuna_hyperparameter_search.cfg
HLasse: You were right - the functionality does not seem to be there already. Created an issue from your comment

psycop/common/model_training_v2/hyperparameter_suggester/optuna_hyperparameter_search.py
HLasse: Hm, the direction is tightly coupled to the metric that we want to optimize which has so far just been AUROC (which should be optimized ;)). 
But yeah, it could be dangerous if someone wants to optimize something else at some point, I'll remove the default

MartinBernstorff: Agreed! Due to the tight coupling, perhaps it should be part of the metrics?

psycop/common/model_training_v2/hyperparameter_suggester/optuna_hyperparameter_search.py
HLasse: Excellent suggestion! Changed to match the regex "^@.*suggester.*", so will resolve if 'suggester' is anywhere in the name of the registry

HLasse: As we just talked about -- agreed, would be extremely neat -- however, not straightforward enough with current architecture. Saving for later

psycop/common/model_training_v2/hyperparameter_suggester/optuna_hyperparameter_search.py
HLasse: As we just talked about -- agreed, would be extremely neat -- however, not high enough priority given the ease of implementation right now. Saving for later

PR #715, opened 2024-01-17
refactor: rename registry to SequenceRegistry

MartinBernstorff:
Makes it easier to separate from the baseline registry. Was indicated for registering the new regionalsplitfilter within a scope accessible by the sequence models (used in #716). 
Hope it's OK with you @KennethEnevoldsen?

I wonder why that's necessary, though. Not sure I quite understand the namespacing of registries. Would love a chat about it if you have time @KennethEnevoldsen.

### DIFF
--- a/psycop/common/feature_generation/sequences/event_loader.py
+++ b/psycop/common/feature_generation/sequences/event_loader.py
@@ -3,7 +3,7 @@ from typing import Protocol, runtime_checkable
 
 import polars as pl
 
-from ...sequence_models.registry import Registry
+from ...sequence_models.registry import SequenceRegistry
 from ..loaders.raw.sql_load import sql_load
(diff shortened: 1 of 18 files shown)


KennethEnevoldsen (commented):
Generally I would say the two shouldn't affect eachother. In the few cases where you might want to register in both I would use import Registry as SequenceRegisty.

It is only a minor preference though. If the office agrees I am on board with this.

MartinBernstorff:
### Merge activity

* **Jan 17, 10:28 AM**: [Graphite](https://app.graphite.dev/github/pr/Aarhus-Psychiatry-Research/psycop-common/715) rebased this pull request after merging its parent, because this pull request is set to merge when ready.

PR #718, opened 2024-01-17
feat: working pretraining

MartinBernstorff:
Running into torch-problems.

image

@KennethEnevoldsen, can you help me out here? ❤️


KennethEnevoldsen:
run in debug mode. Figure out what devices different things are on (tensor.device). Figure out when it happens add some tensor.to(self.device). I imagine it happens in the collate fn

PR #735, opened 2024-01-18
build: update timeseriesflattener

HLasse:
Looks like timeseriesflattener needs Sequence[whatever] instead of List[whatever].


requirements.txt
```diff
@@ -28,10 +28,10 @@ shap==0.41.0
 sqlalchemy==1.4.39
 tableone==0.8.0
 tabulate==0.8.10
-timeseriesflattener==1.8.0
+timeseriesflattener==1.9.0
```
HLasse, line 31:
```suggestion
timeseriesflattener==1.9.1
```

PR #801, opened 2024-01-31
use val y for result metric

sarakolding:


HLasse (approved):
Nice catch! That could easily have flown under that radar and caused a lot of confusion..

MartinBernstorff:
This would've been really bad, thanks a ton for catching it! How did you come about it?

sarakolding:
I think we would have noticed, since we rarely would have train and val sets of equal size - so it would throw an error. Used our pipe for the algorithmic fairness course - it was great! 🐒

Issue #805, opened 2024-02-06
refactor: figure out how to do estimators


MartinBernstorff:
Currently, using an sklearn pipeline gives a bunch of limitations. We cannot:
* Log to the loggers
* Transform the datasets (feature selection)

Some things would be neat, but can also be done in evaluation:
* Access the attributes of the pipeline elements (e.g. feature importances)

However, being able to just save a scikit-learn pipe is also very nice! So, what do we do? Can we create a pipeline-like element, e.g. a subclass of the scikit-learn estimators, which we can use?

HLasse:
One partial solution is to use the [pipeline](https://imbalanced-learn.org/stable/references/generated/imblearn.pipeline.Pipeline.html) from [imbalanced-learn](https://imbalanced-learn.org/stable/) instead of sklearn's pipeline. As previously discussed, it's a drop-in replacement, but allows for estimators that transform the dataset (adding/removing rows - uncertain about cols). 

For feature selection (dropping columns), it seems there are some fairly simple solutions here: https://stackoverflow.com/questions/68402691/adding-dropping-column-instance-into-a-pipeline

Happy to discuss this more!

HLasse:
Also, making own estimators is fairly straightforward. Here's an example of wrapping an existing one (which could be done to e.g. add a logger to an existing pipeline): https://github.com/Aarhus-Psychiatry-Research/psycop-common/pull/607/commits/4be5abc9a1ee7ac67279b51b7f6e32058648484b

HLasse:
Just remembered that I made an estimator for textdescriptives a long time ago - not daunting at all: https://github.com/HLasse/TextDescriptives/blob/main/src/textdescriptives/integrations/sklearn_featurizer.py

PR #806, opened 2024-02-06
feat: feature reduction for sklearn pipeline

MartinBernstorff:
An example of the need for #805.


psycop/common/model_training_v2/trainer/task/binary_classification/binary_classification_pipeline.py
```diff
@@ -16,6 +16,14 @@ class BinaryClassificationPipeline(BasePipeline):
 
     def fit(self, x: pd.DataFrame, y: pd.Series) -> None:  # type: ignore
         self.sklearn_pipe.fit(X=x, y=y)
+        print(
```
MartinBernstorff, line 19: This should never be merged, and is what we need to design around.

PR #809, opened 2024-02-07
feat: extract metrics from mlflow 

HLasse:
Wanted to create a cute little table of performances from my text experiment, but turned out to be more painful than expected to interface with mlflow. Created this little class to help. 

Very open for comments, especially on naming and placement in the repo

### DIFF
--- a/psycop/projects/scz_bp/text_experiment/mlflow_data_extraction.py
+++ b/psycop/projects/scz_bp/text_experiment/mlflow_data_extraction.py
@@ -1,4 +1,4 @@
-from typing import Iterable
+from collections.abc import Iterable
 
 import polars as pl
 from mlflow.entities import Run
@@ -64,4 +64,3 @@ if __name__ == "__main__":
     df = MlflowDataExtraction().get_metrics_for_experiment(
         "text_exp", ["all_oof_BinaryAUROC", "within_fold_0_BinaryAUROC"]
     )
-    
--- a/psycop/projects/scz_bp/text_experiment/mlflow_data_extraction.py
+++ b/psycop/projects/scz_bp/text_experiment/mlflow_data_extraction.py
@@ -1,0 +1,67 @@
+from typing import Iterable
+
+import polars as pl
+from mlflow.entities import Run
+from mlflow.store.entities import PagedList
+from mlflow.tracking import MlflowClient
+
+
+class MlflowDataExtraction:
+    def __init__(self) -> None:
+        self.client = MlflowClient(tracking_uri="http://exrhel0371.it.rm.dk:5050")
+
+    def get_all_metrics_for_experiment(self, experiment_name: str) -> pl.DataFrame:
+        """Get the final value of all logged metrics for each run in an experiment.
+        Returns a long df with cols 'run_name', 'metric', and 'value'."""
+        runs = self._get_mlflow_runs_by_experiment(experiment_name=experiment_name)
+        metrics_df = [self._get_all_metrics_for_run(run) for run in runs]
+        return pl.concat(metrics_df).melt(id_vars="run_name", variable_name="metric")
+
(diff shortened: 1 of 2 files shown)


psycop/projects/scz_bp/text_experiment/mlflow_data_extraction.py
```diff
@@ -0,0 +1,66 @@
...
+            pl.lit(run.info.run_name).alias("run_name")
+        )
+
+    def _get_mlflow_experiment_id_from_experiment_name(self, experiment_name: str) -> str:
+        experiment = self.client.get_experiment_by_name(name=experiment_name)
+        if experiment is None:
+            raise ValueError(f"{experiment_name} does not exist on MlFlow.")
+        return experiment.experiment_id
+
+    def _get_mlflow_runs_by_experiment(self, experiment_name: str) -> PagedList[Run]:
```
MartinBernstorff, line 55: What is the consequence of a PagedList here?

psycop/projects/scz_bp/text_experiment/mlflow_data_extraction.py
```diff
@@ -0,0 +1,66 @@
+from collections.abc import Iterable
+
+import polars as pl
+from mlflow.entities import Run
+from mlflow.store.entities import PagedList
+from mlflow.tracking import MlflowClient
+
+
+class MlflowDataExtraction:
```
MartinBernstorff, line 9: `MLFlowMetricExtractor`?

MartinBernstorff:
Looks great! Seems like an annoying problem, with a nice solution.

I'm likely biased, but would love the returntypes to be a subclass of `ValidatedFrame`! 

Then we can read the dataframe specification from the types instead of a docstring, and we ensure it stays up-to-date 👍 

What do you think here?

MartinBernstorff:
We could also test this by mocking the client, but I personally don't find it that important here 👍

HLasse:
Yes, strong agree on the ValidatedFrame - I thought the same, but wanted your thoughts first!

psycop/projects/scz_bp/text_experiment/mlflow_data_extraction.py
```diff
@@ -0,0 +1,66 @@
...
+            self._get_metric_for_run(run_id=run.info.run_id, metric=metric) for metric in metrics
+        ]
+        return pl.concat(metrics_df).with_columns(pl.lit(run.info.run_name).alias("run_name"))
+
+    def _get_metric_for_run(self, run_id: str, metric: str) -> pl.DataFrame:
+        return pl.DataFrame(
+            [
+                dict(metric_object)
+                for metric_object in self.client.get_metric_history(run_id=run_id, key=metric)
+            ]
```
MartinBernstorff, line 41: Just want to highlight an Iterpy solution here:

`return Iter(self.client.get_metric_history(run_id=run_id, key=metric)).map(dict).to_list()`

MartinBernstorff: Huh, I expected that to be much more readable, but it was only OK 🤷‍♂️

psycop/projects/scz_bp/text_experiment/mlflow_data_extraction.py
HLasse: That's the return type of the search function. Doesn't matter for us - changed to iterable

psycop/projects/scz_bp/text_experiment/mlflow_data_extraction.py
HLasse: Hm, yeah - not sure which I prefer tbh 🤷‍♂️

HLasse:
Also added functionality to retrieve the best run across one or more experiments 👍

PR #820, opened 2024-02-20
fix deprecated parsing

sarakolding:
realised that newer versions of polars have changed how strings are parsed in when-then statements (literals vs columns, see https://github.com/pola-rs/polars/pull/10122)
updated where applicable (also in your repos, @MartinBernstorff and @HLasse)


MartinBernstorff:
Love this! Thanks a ton Sara ❤️ 

Exactly the benefit of having shared code.

MartinBernstorff (approved):

PR #861, opened 2024-03-11
feat: unify naming classification model steps

HLasse:
This ensures that when we want to extract e.g. feature importance from the models, they will always be found under `pipeline.named_steps["classifier"]` instead of being the name of the model. Requires us to follow the naming convention when/if adding additional classification models, but no further downsides as I see it


MartinBernstorff:
This looks great! Agreed on your considerations; i.e., I'm not sure the tuple is even needed, but IIRC it's expected by sklearn?

MartinBernstorff (approved):

HLasse:
Sweet, merging! Yep, the tuple is expected by sklearn

PR #874, opened 2024-04-02
change name of reqs file to requirements

sarakolding:
snyk looks for *requirements.txt pattern, so this file was not scanned for vulnerabilities


sarakolding:
@HLasse is it on purpose you haven't named it similarly to the other req files?

HLasse:
Nope, unintended

sarakolding:
hmm, i think maybe the dev container uses the same pattern, so because the files have different versions of mlflow, there are conflicting dependencies when trying to set up the container

HLasse:
I can clean up the synthcity-reqs to only be the diff to the main dependencies - that should make our lives easier

sarakolding:
that would be 🥚cellent

HLasse:
On further thought, there was a reason ;) the synthcity-reqs are incompatible with some functionality in the rest of the library (hence the separate file and not just added to the core requirements), which means that there will be conflicts and some tests will fail if they are installed together. 

I have manually checked the synthcity-reqs for vulnerabilities before installing on Overtaci. Perhaps the synthcity-reqs file should be moved into the scz-bp project folder, so no-one uses it "on accident" - its only purpose is for testing models that are augmented with synthetic data.

sarakolding:
alrigthy 👍

PR #880, opened 2024-04-16
perf: benchmark new timeseriesflattener

MartinBernstorff:
@sarakolding


sarakolding:
@MartinBernstorff do you want to keep this?

MartinBernstorff:
Nah, this is quite chaotic. It is probably a good idea for us to build a proper benchmark at some point, if we keep needing to optimise, but let's defer it once more.

PR #893, opened 2024-04-30
aggressive stalebot

sarakolding:
i feel like the stalebot is a little too aggressive. any strong feelings about this @MartinBernstorff @HLasse ?


MartinBernstorff:
Interesting! 

I think we set the thresholds based on the idea that PRs should be merged rapidly. If they take a long time to merge, it indicates that their scope was too large.

But that might be wrong, or we might have missed something! When have you found it to be too aggressive?

MartinBernstorff:
@sarakolding It seems like we forgot about this. Would you like to talk about it?

sarakolding:
@MartinBernstorff no, I think you're right

Issue #907, opened 2024-05-15
fix: config .cfg is not being logged?


MartinBernstorff:
E.g. http://exrhel0371.it.rm.dk:5050/#/experiments/512442793765203216/runs/31d5a58d1ac143b9bda885e737fd3069

@HLasse Same problem with your runs? Have we modified anything here recently?

PR #908, opened 2024-05-15
refactor: ensure logging when training from mutated cfg

MartinBernstorff:
@HLasse Post-merge review. When programatically changing configs, recommend we use `train_baseline_model_from_cfg`. Was the cause of my confusion in #907.


HLasse (commented):
Nice change 👍

PR #911, opened 2024-05-15
feat: mlflow query for run and get eval_df

MartinBernstorff:
Small change, hope it's OK with a check for style :-)


HLasse (commented):
Neat!

PR #913, opened 2024-05-16
model performance confusion matrix for sczbp

sarakolding:


sarakolding:
requires test?

sarakolding:
function take df instead of mlflow run. move run to project repo

PR #962, opened 2024-06-18
feat: feature description tsflattener v2

### LINKED ISSUE #960: Feature description table not compatible with tsflattener v2
We can no longer simply supply the specs, as each spec can be multiple features. Instead, we should probably parse the column names and extract the relevant information from them.

HLasse:
Fixes #960 

Lots of type ignores because 1) `pl.mean` can return None (??), and a regex match can both be a match object and None..


psycop/common/feature_generation/data_checks/flattened/feature_describer_tsflattener_v2.py
```diff
@@ -0,0 +1,98 @@
...
+def tsflattener_v2_column_is_static(col_name: str) -> bool:
+    str_match = "within"
+    return str_match not in col_name
+
+
+def get_value_proportion(series: pl.Series, value: str) -> float:
+    """Get proportion of series that is equal to the value argument."""
+    if value == "nan":
+        return round(series.is_nan().mean(), 2)  # type: ignore
+    return round(series.eq(float(value)).mean(), 2)  # type: ignore
```
sarakolding: what if fallback value is also a "naturally occurring" value in the data?

psycop/common/feature_generation/data_checks/flattened/feature_describer_tsflattener_v2.py
```diff
@@ -0,0 +1,98 @@
...
+        mean = round(df[col].drop_nans().mean(), 2)  # type: ignore
+        proportion_using_fallback = get_value_proportion(df[col], parsed_col.fallback)
+
+        feature_description = {
+            "Feature name": parsed_col.feature_name,
+            "Lookbehind period": parsed_col.time_interval_start
+            if parsed_col.is_static
+            else f"{parsed_col.time_interval_start} to {parsed_col.time_interval_end} {parsed_col.time_interval_format}",
+            "Resolve multiple": parsed_col.resolve_multiple_strategy,
+            "Fallback strategy": parsed_col.fallback,
```
sarakolding, line 82: maybe a bit off topic but i just wondered whether "fallback strategy" is something we came up with or a more general naming convention? i would tend to think of this "imputation value"

psycop/common/feature_generation/data_checks/flattened/feature_describer_tsflattener_v2.py
```diff
@@ -0,0 +1,98 @@
...
+
+def generate_feature_description_df(
+    df: pl.DataFrame,
+    column_name_parser: Callable[[str], ParsedPredictorColumn] = parse_predictor_column_name,
+) -> pl.DataFrame:
+    feature_rows = []
+    for col in df.columns:
+        parsed_col = column_name_parser(col)
+        n_unique = df[col].n_unique()
+        proportion_missing = round(df[col].is_nan().mean(), 2)  # type: ignore
```
sarakolding, line 72: so this is only if we chose nan as fallback, so functionally the same as "proportion using fallback" for those cols? and otherwise always 0%? seems a bit redundant in that case - maybe more interesting how many were missing before fallback imputation? but i guess that's what the "proportion using fallback" does?

psycop/common/feature_generation/data_checks/flattened/test_feature_describer.py
```diff
@@ -0,0 +1,90 @@
...
+            "pred_female_fallback_nan": [0, 1, np.nan],
+            "outc_first_scz_or_bp_within_0_to_730_days_max_fallback_0": [0, 0, 1],
+            "pred_value_1_within_0_to_730_days_mean_fallback_nan": [1, 2, 3],
+        }
+    )
+
+    expected = pl.DataFrame(
+        {
+            "Feature name": ["female", "first_scz_or_bp", "value_1"],
+            "Lookbehind period": ["N/A", "0 to 730 days", "0 to 730 days"],
```
sarakolding, line 64: what's the diff between nan and n/a?

sarakolding (approved):
code looks great to me - just a few clarifying questions for me, since it's the first time i use this functionality

sarakolding:
whoops forgot i should suggest change instead of just committing my suggestions (couldn't figure out how to open pr in vscode)

psycop/common/feature_generation/data_checks/flattened/feature_describer_tsflattener_v2.py
HLasse: we can't protect against this without seeing the raw data as well. I think it'll have to be a note in the table description

psycop/common/feature_generation/data_checks/flattened/feature_describer_tsflattener_v2.py
HLasse: We came up with fallback strategy. I agree that this is basically imputation, but perhaps it might be conflated with the other type of imputation that is (optionally) done during preprocessing? We use fallback in tsflattener, so might be good to be consistent?

psycop/common/feature_generation/data_checks/flattened/feature_describer_tsflattener_v2.py
HLasse: Yeah, let's remove proportion missing, it's redundant

psycop/common/feature_generation/data_checks/flattened/test_feature_describer.py
HLasse: N/A = not applicable, nan = Not a number

psycop/common/feature_generation/data_checks/flattened/test_feature_describer.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+    assert tsflattener_v2_column_is_static("pred_female_fallback_nan") is True
+
+
+def test_generate_feature_description_df():
+    df = pl.DataFrame(
+        {
+            "pred_female_fallback_nan": [0, 1, np.nan],
+            "outc_first_scz_or_bp_within_0_to_730_days_max_fallback_0": [0, 0, 1],
+            "pred_value_1_within_0_to_730_days_mean_fallback_nan": [1, 2, 3],
```
MartinBernstorff, line 57: Nitpick: Would parametrise these, testing one column at a time. Makes the "expected" df easier to read.

psycop/common/feature_generation/data_checks/flattened/test_feature_describer.py
```diff
@@ -0,0 +1,77 @@
...
+        }
+    )
+
+    expected = pl.DataFrame(
```
MartinBernstorff, line 61: `str_to_pl_df` instead of dict?

psycop/common/feature_generation/data_checks/flattened/feature_describer_tsflattener_v2.py
```diff
@@ -0,0 +1,89 @@
...
+
+def parse_predictor_column_name(
+    col_name: str,
+    is_static: Callable[[str], bool] = tsflattener_v2_column_is_static,
+    feature_name_regex: str = r"[a-zA-Z]+_(.*?)_(?=fallback|within)",
+    fallback_regex: str = r"_fallback_(.+)$",
+    time_interval_start_regex: str = r"_within_([0-9]+)",
+    time_interval_end_regex: str = "_to_([0-9]+)_",
+    time_interval_format_regex: str = r"to_[0-9]+_([a-z]+)_",
+    resolve_multiple_strategy_regex: str = r"([a-z]+)_fallback",
```
MartinBernstorff, line 33: Haha, you sure went all-in on dependency injection/parametrisation here! ❤️ 

Makes the code super-reusable, even if we end up changing the format again ❤️

psycop/common/feature_generation/data_checks/flattened/feature_describer_tsflattener_v2.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def parse_predictor_column_name(
+    col_name: str,
+    is_static: Callable[[str], bool] = tsflattener_v2_column_is_static,
+    feature_name_regex: str = r"[a-zA-Z]+_(.*?)_(?=fallback|within)",
+    fallback_regex: str = r"_fallback_(.+)$",
+    time_interval_start_regex: str = r"_within_([0-9]+)",
+    time_interval_end_regex: str = "_to_([0-9]+)_",
+    time_interval_format_regex: str = r"to_[0-9]+_([a-z]+)_",
```
MartinBernstorff, line 32: Nitpick: Time-interval granularity?

psycop/common/feature_generation/data_checks/flattened/feature_describer_tsflattener_v2.py
```diff
@@ -0,0 +1,89 @@
...
+    resolve_multiple_strategy_regex: str = r"([a-z]+)_fallback",
+) -> ParsedPredictorColumn:
+    col_is_static = is_static(col_name)
+    feature_name = re.search(feature_name_regex, col_name).group(1)  # type: ignore
+    fallback = re.search(fallback_regex, col_name).group(1)  # type: ignore
+    time_interval_start = (
+        re.search(time_interval_start_regex, col_name).group(1) if not col_is_static else "N/A"  # type: ignore
+    )
+    time_interval_end = (
+        re.search(time_interval_end_regex, col_name).group(1) if not col_is_static else "N/A"  # type: ignore
```
MartinBernstorff, line 42: I've submitted a PR cleaning this up a bit and removing the type-errors: #964

psycop/common/feature_generation/data_checks/flattened/feature_describer_tsflattener_v2.py
MartinBernstorff: Also, some of the fallbacks are not estimated from the data (e.g. the NaN), so they are technically not imputed until preprocessing.

psycop/common/feature_generation/data_checks/flattened/test_feature_describer.py
```diff
@@ -0,0 +1,77 @@
...
+    parsed_col = parse_predictor_column_name(predictor_column_example.col_name)
+    assert parsed_col == predictor_column_example
+
+
+def test_tsflattener_v2_is_static():
+    assert (
+        tsflattener_v2_column_is_static("pred_value_1_within_0_to_730_days_mean_fallback_nan")
+        is False
+    )
+    assert tsflattener_v2_column_is_static("pred_female_fallback_nan") is True
```
MartinBernstorff, line 49: This is probably testing implementation instead of functionality, so I would remove it.

psycop/common/feature_generation/data_checks/flattened/test_feature_describer.py
```diff
@@ -0,0 +1,77 @@
...
+            time_interval_end="N/A",
+            time_interval_format="N/A",
+            resolve_multiple_strategy="N/A",
+            is_static=True,
+        ),
+    ],
+)
+def test_parse_predictor_column_name(predictor_column_example: ParsedPredictorColumn):
+    parsed_col = parse_predictor_column_name(predictor_column_example.col_name)
+    assert parsed_col == predictor_column_example
```
MartinBernstorff, line 41: Similar to the below, since "parse_predictor_column_name" could be a private function. Would like to discuss 👍

MartinBernstorff (commented):
Great work Lasse, super nice to have this fixed up!

A few nitpicks, and a refactor in another PR :-)

Maybe this should be part of timeseriesflattener, actually? They are tightly coupled, and this could be very nice for other users as well.

PR #981, opened 2024-07-12
feat: PsycopConfig

MartinBernstorff:
Adds a PsycopConfig wrapper around `confection.Config`.

Advantages:
* Mutate method, which errors if the attribute you are indexing to does not exist. Previously, if you mis-typed the last index, it would create a new key (a silent failure on typo)
* Retrieval gives a nicer error than the flat keyerror from a dict
* Shorthand for indexing which matches the shorthand in the `.cfg` files

Disadvantages:
* An extra layer of indirection

I'll start using this, so it's a post-merge review and an FYI :-)


psycop/common/global_utils/cache.py
```diff
@@ -8,7 +8,7 @@
 if Path("E:/").exists():
     cache_dir = OVARTACI_SHARED_DIR / "cache"
 else:
-    cache_dir = PSYCOP_PKG_ROOT / "test_utils" / "test_outputs" / "sql_cache"
+    cache_dir = PSYCOP_PKG_ROOT / ".cache"
```
MartinBernstorff, line 11: Makes the local cache dir excluded from git

PR #992, opened 2024-08-08
chore: pin matplotlib version

MartinBernstorff:


MartinBernstorff:
Force merging because only vuln scan is failing. This enables #985 to merge.

PR #994, opened 2024-08-12
feat: cvd test-set training

MartinBernstorff:
@HLasse, would love a check of both:
* The logic (am I missing anything for training on the test set?)
* The new addition to PsycopConfig (allowing for a fluent interface, e.g. chained method calls)

I've renamed the methods for easier comparison of the location strings. Since there are only 4 methods, I think the trade-off is OK, but let me know what you think 👍 

Thanks!


psycop/projects/cvd/model_training/eval_test_set.py
```diff
@@ -0,0 +1,48 @@
+from pathlib import Path
+
+from psycop.common.model_training_v2.config.baseline_pipeline import train_baseline_model_from_cfg
+from psycop.common.model_training_v2.config.config_utils import PsycopConfig
+from psycop.common.model_training_v2.config.populate_registry import populate_baseline_registry
+from psycop.projects.cvd.model_training.populate_cvd_registry import populate_with_cvd_registry
+
+
+def test_set(cfg: PsycopConfig):
```
HLasse, line 9: I love this

HLasse: Avoids so much pain

PR #1009, opened 2024-09-11
fix: `QuarantineFilter` does not remove prediction times on same day

### LINKED ISSUE #1008: QuarantineFilter does not remove cases earlier in the same day
image

HLasse:
Fixes #1008 

Note @bokajgd @erikperfalk @sarakolding if you have used `QuarantineFilter` to remove e.g. incident cases in the last X years, if the outcome happened on the same day as a prediction time they were not filtered. Probably not major (in my case, approximately 200 prediction times that should have been removed)


MartinBernstorff:
Ah, this is a very good fix! Looks good 👍

PR #1024, opened 2024-09-24
add fairness package

sarakolding:


sarakolding:
tried to add a fairness package, but it clashed with our pandas version, so I had to update. also had to update timeseriesflattener, because it had an upper bound for pandas. so some changes are related to updating tsflattener (e.g., that the new version no longer supports lazy computing), while others are related to changed functionality in pandas (e.g,. in casting strings to timestamps).

sarakolding:
had some weird errors that i couldn't really figure out - e.g., we are suddenly way more sensitive to the number of decimals in seconds in timestamps. i just changed it to fit, so i hope it was not part of some test of functionality (since we can't check with the original code authors anymore 🥲)

sarakolding:
@bokajgd review my pr plz 🥺

bokajgd (approved):
Sorrryyyy, ved ikke hvordan jeg kom fra dette igen. Det ser super fint ud! :)

PR #1071, opened 2025-01-28
scz_bp rerun

sarakolding:
changes made to be able to rerun results, including loading eval ds from disk and adding disk logger


sarakolding:
@bokajgd need the load_from_disk function for the fairness demo, so if you could just take a quick look that everything looks okay, that would be great 😇

bokajgd (approved):
Very nice!

PR #1087, opened 2025-05-14
feat: add new f9 loaders

sarakolding:
as requested, added two new loaders: one for loading all f9* disorders except f99* and one loading only f99*. kept the old one for replication purposes, but we should only use the new ones going forward. maybe we should add a comment or a log message detailing this when people try to use the old loader?

additionally, i realised that the "exclude_codes" arg in the from_codes() loader **does not work for diagnoses**, only atc codes, since it assumes that there is only one code per row. i have tried to make this more explicit by naming the arg "exclude_atc_codes" instead. i only found one place where it was used for diagnoses, but it was in a subfolder called loaders_2024, so i assume this is deprecated, and not worth changing.

### DIFF
--- a/psycop/common/feature_generation/loaders_2024/diagnoses.py
+++ b/psycop/common/feature_generation/loaders_2024/diagnoses.py
@@ -157,6 +157,7 @@ def load_from_codes(
     output_col_name: str | None = None,
     match_with_wildcard: bool = True,
     n_rows: int | None = None,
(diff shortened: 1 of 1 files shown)


bokajgd (changes requested):
Everything looks super! Where was it used in the loaders_2024 folder? These loaders are not necessarily deprecated - they just load in data up to 2024 instead of 2021 for the projects that have access to this data.

PR #1089, opened 2025-05-20
fix: lookbehind combination col filter

sarakolding:
when selecting a combination of lookbehinds, we were selecting only based on integers - meaning if you were trying to select predictors with lookbehinds of 1, you would not only get within_1_days, but also within_10_days or even f1_disorders


bokajgd (approved):
What a catch! Ser super fint ud.

PR #1127, opened 2025-09-26
add stratify arg to auroc

sarakolding:


sarakolding:
can't remember what i thought i had forgotten

